PR #31, opened 2014-06-20
Fix formfield to return an empty string if an empty value is given. This...

bramd:
... allows empty input for not null model fields with blank=True.


amateja:
It is fixed but in slightly different way. Could you retest it if it is OK for you now?

bramd:
Looks fine to me, thanks.

amateja:
Superb! Thank you.

PR #37, opened 2014-07-23
allow save null values to phonenumber field

GeyseR:
Hi!

This PR fixes regression introduced in v0.6. 
In my model I want to phone be empty and unique if it's not empty. Since last changes in https://github.com/stefanfoulis/django-phonenumber-field/commit/1c7b803121145098d85bb6ab547657ab15519862 all emptye values (None and '') saves as empty strings, but database interpret empty strings as NOT NULL values and raises IntegrityError.

Old tests not broken and new added.

BTW, thanks for great app! Very useful for me!


amateja:
It should be fixed now... but with a different PR. Could you retest if everything works OK for you?

PR #42, opened 2014-08-16
Fix #41

### LINKED ISSUE #41: Can't run South migration to add a PhoneNumberField
I'm adding a PhoneNumberField field (with blank=True) to an existing model. As such, when I run `./manage schemamigration --auto myapp`, South demands that I provide a default value for the column it will be adding. If I set the default (either in models.py or as a one-off in the South migration) to `""` (an empty string), when I try to run the migration I get an AttributeError:

`AttributeError: 'NoneType' object has no attribute 'as_e164'`

This is because in [`get_prep_value`](https://github.com/stefanfoulis/django-phonenumber-field/blob/develop/phonenumber_field/modelfields.py#L55), the first if block is passed (since `""` does not equal None) and the `to_python` function changes value from "" to `None` so when [line 67](https://github.com/stefanfoulis/django-phonenumber-field/blob/develop/phonenumber_field/modelfields.py#L67) is hit, we get the AttributeError.

If I try setting the default to `None`, I then get an IntegrityError:

`django.db.utils.IntegrityError: column "phone_number" contains null values`

As the Django docs [recommend](https://docs.djangoproject.com/en/dev/ref/models/fields/#null) and as mentioned in [issue 1](https://github.com/stefanfoulis/django-phonenumber-field/issues/1#issuecomment-2105919), we should use an empty string, not null, to represent a blank value for a CharField (which PhoneNumberField is under the hood). As such, although adding `null=True` to the field definition in my models makes the migration succeed, it leads to trouble down the road as you can end up with both `""` and `None` as null values.

Not sure exactly how or where you'd like to change things but the request is that it be possible to have South work properly when defining a PhoneNumberField with blank=True. Let me know if I'm misunderstanding something here. Thanks!

jeffbowen:
This seems to solve #41. Let me know if my understanding of how this should work is wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,15 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
 install:
   - ./.travis_setup
 env:
-  - DJANGO="django>=1.4,<1.5"
   - DJANGO="django>=1.5,<1.6"
   - DJANGO="django>=1.6,<1.7"
+  - DJANGO="django>=1.7,<1.8"
 script:
-  - cd ./testproject; python manage.py test phonenumber_field testapp
-matrix:
-  exclude:
-    - python: "3.3"
-      env: DJANGO="django>=1.4,<1.5"
-    - python: "3.4"
-      env: DJANGO="django>=1.4,<1.5"
+  - cd ./testproject; coverage run --source=phonenumber_field manage.py test phonenumber_field testapp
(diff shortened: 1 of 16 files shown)


vpatel85:
:+1: can you merge this PR?

thedrow:
We need @jeffbowen to rebase because we can't merge this as it is.

jeffbowen:
Okay, good to merge now @thedrow.

PR #45, opened 2014-09-17
using negotiate method

zauddelig:
I propose this because It ignores case


amateja:

PR #49, opened 2014-10-17
Let python-phonenumbers compare PhoneNumber instances

cormier:
in `phonenumber.py`, the overridden `__eq__` converts the number to e_164 before the comparison:
```python
def __eq__(self, other):
    if type(other) == PhoneNumber:
        return self.as_e164 == other.as_e164
    else:
        return super(PhoneNumber, self).__eq__(other)
```
It seems that the e_164 conversion does not include the extension, and it causes a comparison of two PhoneNumbers to be erroneous when only the extension is divergent. For example:
```python
> other
PhoneNumber(country_code=1, national_number=8005551234, extension='2', italian_leading_zero=None, number_of_leading_zeros=None, country_code_source=20, preferred_domestic_carrier_code='')
> self
PhoneNumber(country_code=1, national_number=8005551234, extension=None, italian_leading_zero=None, number_of_leading_zeros=None, country_code_source=20, preferred_domestic_carrier_code='')
> other == self
True
```
In my opinion, `django-phonenumber-field` should not be involved in the comparison, and should delegate all operations that involve phone number logic to `python-phonenumbers`.

### DIFF
--- a/phonenumber_field/phonenumber.py
+++ b/phonenumber_field/phonenumber.py
@@ -75,12 +75,6 @@ class PhoneNumber(phonenumbers.phonenumber.PhoneNumber):
     def __len__(self):
         return len(self.__unicode__())
 
-    def __eq__(self, other):
-        if type(other) == PhoneNumber:
-            return self.as_e164 == other.as_e164
-        else:
-            return super(PhoneNumber, self).__eq__(other)
-
 
 def to_python(value):
     if value in validators.EMPTY_VALUES:  # None or ''
--- a/phonenumber_field/tests.py
+++ b/phonenumber_field/tests.py
@@ -3,6 +3,8 @@
 from django.test.testcases import TestCase
 from django.db import models
 
+from phonenumbers import is_number_match, MatchType
+
 from phonenumber_field.modelfields import PhoneNumberField
 from phonenumber_field.phonenumber import PhoneNumber
 from phonenumber_field.validators import to_python
@@ -52,7 +54,8 @@ class PhoneNumberFieldTestCase(TestCase):
             MandatoryPhoneNumber.objects.create(
                 phone_number=number_string).phone_number
             for number_string in self.equal_number_strings]
-        self.assertTrue(all(n == numbers[0] for n in numbers))
+        self.assertTrue(all(is_number_match(n, numbers[0]) == MatchType.EXACT_MATCH
+                        for n in numbers))
 
     def test_field_returns_correct_type(self):
         model = OptionalPhoneNumber()


thedrow:
Can you please rebase?

cormier:
With pleasure, @thedrow !

thedrow:
It seems to me that while this is a good change it's also a breaking change.
It might break someone else's code.
What do you think?

cormier:
I agree that it could break some code. I still believe that it's the right thing to do, and that our approach should be to mitigate the risks of unexpected side effects. The changes that are proposed are minimal, and I believe that we should be able to identify how it may break code and document these changes in the release notes so people can take notice when upgrading the module.

One obvious way that it will break the code is for people who _expect_ the extension to not be taken into account by the comparison:
```python
>>> PhoneNumber.from_string('1-514-555-1234', region='US') == PhoneNumber.from_string('1-514-555-1234 ext. 122', region='US')
True
```

Other than this, comparing the `phonenumberfield.phonenumber.PhoneNumber` to a `phonenumbers.phonenumber.PhoneNumber` still works. For example:
```python
>>> ph = PhoneNumber.from_string('15145551234', region='US')
>>> ph2 = phonenumbers.parse(number='15145551234', region='US', keep_raw_input=True)
>>> ph == ph2
```

I believe that other ways that it may break the code relates to information that is « hidden » by the E.164 format and used in the `__eq__` comparator of `phonenumbers.phonenumber.Phonenumber`.

`phonenumbers.phonenumber.PhoneNumber`'s `__eq__`[1] method reads like this:

```python
    def __eq__(self, other):
        if not isinstance(other, PhoneNumber):
            return False
        return (self.country_code == other.country_code and
                self.national_number == other.national_number and
                self.extension == other.extension and
                bool(self.italian_leading_zero) == bool(other.italian_leading_zero) and
                self.number_of_leading_zeros == other.number_of_leading_zeros and
                self.raw_input == other.raw_input and
                self.country_code_source == other.country_code_source and
                self.preferred_domestic_carrier_code == other.preferred_domestic_carrier_code)
```
I believe that comparison will "break" for everyone who expect these values *not* to be compared:
* `italian_leading_zero`
* `number_of_leading_zeros`
* `raw_input`
* `country_code_source`
* `preferred_domestic_carrier_code`

In my opinion, this is a "fair" price to pay to maintain the behavior that users of python-phonenumbers will expect. However, I do not know what motivated implementing the E.164 comparison in the first place. Maybe we should keep it in a separate method (e.g. `simple_compare`) and ask users to move to this one?

References:
[1] https://github.com/daviddrysdale/python-phonenumbers/blob/0e7d0dacfff20dd721e74c13b59e2410ca7a4571/python/phonenumbers/phonenumber.py#L194

thedrow:
At this point I'd like to ask @stefanfoulis what he thinks.

thedrow:
Since @stefanfoulis did not respond at all I'm going to merge this.

PR #54, opened 2014-11-07
Fixing a minor bug.

pegasus55:
When using PhoneNumberPrefixWidget and leaving the field empty, it doesn't save a '.' string anymore.
Also, if the prefix select or the textfield is empty, it will not save the field.


thedrow:
Can you please rebase?

amateja:

PR #58, opened 2015-01-20
Fix issue with language codes that include territory

jiaaro:
`translation.get_language()` returns "en-us" in my development environment. That does not work as is, but this pull request fixes it :)

Also, added Babel to the dependencies since the module doesn't work without it


amateja:

PR #61, opened 2015-03-01
fixed setup dependency on python-babel and fixed widget local bug

siolag161:
The django method `get_language` returns a string of format "LANG-COUNTRY" whereas babel Locale needs an locale format  "LANG_COUNTRY". 

I also added the dependency on the package babel

### DIFF
--- a/phonenumber_field/formfields.py
+++ b/phonenumber_field/formfields.py
@@ -9,6 +9,10 @@ from phonenumber_field.validators import validate_international_phonenumber
 from phonenumber_field.phonenumber import to_python
 
 
+import logging
+logger = logging.getLogger('werkzeug')
+
+
 class PhoneNumberField(CharField):
     default_error_messages = {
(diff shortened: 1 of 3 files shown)


EugeneDae:
This is what I get when I launch the dev server with this fork:

    (prj)Dae-MBP:src Dae$ ./manage.py runserver
    13:11:19 CRITICAL None
    13:11:19 CRITICAL None

The fix is to remove line 21 `logger.critical(attrs)` in widgets.py

siolag161:
Actually I **do not** think it's a good idea to merge it.

There are 2 commits in total, [one of which fixed the bug related to Babel](https://github.com/siolag161/django-phonenumber-field/commit/65d50303b60b2ded74146434281298242bcd9884) and it's a simple fix as you can see, whereas the other was to satisfy my own needs.

I was careless not to remove this commit and the log thing as I am new to this.

amateja:

PR #38, opened 2014-07-28
Add PHONENUMBER_SERIALIZE_FORMAT setting

hsoft:
The E164 format doesn't include extensions and if someone wants to keep
this information, he has to either create a new field or change the
serialization format.

This commit allows us to change our serialization format through the
PHONENUMBER_SERIALIZE_FORMAT setting.

PR #63, opened 2015-03-16
Support saving phone numbers with extension in rfc3966 format

huandzh:
Hi, I had run into a problem that extensions were not stored in database properly. In the proposed pull request, a new argument `number_format` supporting both  `e1641` and `rfc3966` was added to address it.

Currently, it service well for me. But I am not familiar with phone number formats. 

Any ideas? THX

### DIFF
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -40,11 +40,15 @@ class PhoneNumberField(models.Field):
     attr_class = PhoneNumber
     descriptor_class = PhoneNumberDescriptor
     default_validators = [validate_international_phonenumber]
+    writing_formats = ['e164','rfc3966']
 
     description = _("Phone number")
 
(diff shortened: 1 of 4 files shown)


hsoft:
Kind of related to my own PR, #38...

huandzh:
Yep, they targeted almost the same issue.

I preferred argument per field, in case existing mobile phone numbers were using `e_164`.  And i am also not sure whether `international` and `national` are among interchangeable phone number formats.

huandzh:
Yep, they targeted almost the same issue.

I preferred argument per field, in case existing mobile phone numbers were using `e_164`. 

And i am also not sure whether `international` and `national` are interchangeable phone number formats.

tdobrovolny:
This is solved by already accepted pull request #75 . You can choose DB storage format which you favour.

thedrow:
Thanks for the heads up!

PR #68, opened 2015-04-18
Fix field validation failing on no number provided by user

Informatic:
Validation in `PhoneNumberField.to_python` fails when number is not provided (even if it is not required) because `PhoneNumberPrefixWidget.value_from_datadict` returns `r'.'`.
This fix makes it return `None` if user hasn't selected anything.
When field is required everything works as before (just fails on "Field is required" instead)

Someone might want to make some tests against this case, i don't really feel like it right now... :P


thedrow:
I'm not inclined to merge this without tests.
I'll keep this PR open until you or anyone else will provide them.

amateja:

PR #74, opened 2015-06-17
Return value based on empty type

tredzko:
Currently, when a phone number is `NULL` or it is an empty string, the returned value from `to_python()` is `None`. 

**Steps to reproduce**
1. Use the `PhoneNumberField` model to create a phone number with `blank=true` and `default=''`.
2. Get the value of the phone number.
3. The value will be `None` instead of a blank string.

This can be a bit of a headache when it comes to using an already-entered value in an editable field.

This change allows the phone number to be pulled from the database as a blank string if it's blank, or as `None` if it's value is `NULL` in the database.

This request also updates the appropriate tests regarding this change.

### DIFF
--- a/phonenumber_field/phonenumber.py
+++ b/phonenumber_field/phonenumber.py
@@ -78,7 +78,7 @@ class PhoneNumber(phonenumbers.phonenumber.PhoneNumber):
 
 def to_python(value):
     if value in validators.EMPTY_VALUES:  # None or ''
-        phone_number = None
+        phone_number = value
     elif value and isinstance(value, string_types):
         try:
             phone_number = PhoneNumber.from_string(phone_number=value)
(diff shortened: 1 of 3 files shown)


kevin-brown:
Might as well mention that when the field is initialized with `null=False` (as recommended), the current behaviour is to still return `None` when the field is empty, which is almost equivalent to `null=True`.

This fixes that behaviour and brings it back in-line with the standard `CharField`, which just returns the blank string if one is saved in the database.

kevin-brown:
@thedrow Is there any chance you can review this pull request?

Right now the field returns `None` even when `null=False` is set.

tdobrovolny:
I am not sure, that it is correct behavior for PhoneNumberField. When we look at the PhoneNumberField as CharField, so definitly it is correct behavior . But in my opinion PhoneNumberField is not CharField. It is "internaly" (for our convenience and bigger compatibility), when store to database come to place, CharField, but I think  we should implement behavior as IntegerField, FloatField BooleanField, DateField, DateTimeField ... These Field types use as to_python function this behavior:
```python
if value is None:
    return value:
try:
    return fieldtype()
except (TypeError, ValueError)
    raise exceptions.ValidationError(...
````
So in my opinion for empty and None values, our to_python should return None. For nonempty strings we should return PhoneNumber or raise validation error.

Please discuss this!

Issue #73, opened 2015-06-05
Extensions are silently stripped


eswald:
Saving a value like "424-417-2469x61914" results in the value "+14244172469", with the extension removed, instead of either saving the extension or throwing an error.

thedrow:
Resolved in #75.

PR #75, opened 2015-07-04
Configurable string representation of phone number in database backend

tdobrovolny:
This pull request allow developer to choose his optimal string representation of phone number in database backend.

Configuration is done through django settings variable PHONENUMBER_DB_FORMAT set to one of defined string representations in PhoneNumber.format_map. Default value, when variable is unset is E164, so this modification is fully compatible with already installed instances, which does not recognize any changes.

Moreover it is possible (but not recommended) to change database representation during project live, because reverse conversion was originally done by from_string class method which use parse method which correctly interpret all standard formats (national format needs to setup PHONENUMBER_DEFAULT_REGION).

This pull request solve issue #73 - Extensions are silently stripped because developer can set RFC3966 format, which preserve information about extension instead of E164, which strip out extension info.


phonenumber_field/modelfields.py
```diff
@@ -65,7 +66,9 @@ def get_prep_value(self, value):
         if isinstance(value, string_types):
             # it is an invalid phone number
             return value
-        return value.as_e164
```
tdobrovolny: Fixed E164 string representation

phonenumber_field/modelfields.py
```diff
@@ -65,7 +66,9 @@ def get_prep_value(self, value):
...
+        format_string = getattr(settings, 'PHONENUMBER_DB_FORMAT', 'E164')
```
tdobrovolny: Get configured phone number db string representation with default E164

phonenumber_field/tests.py
```diff
@@ -34,6 +35,11 @@ class PhoneNumberFieldTestCase(TestCase):
         ('GB', '01606 751 78'),
         ('DE', '0176/96842671'),
     ]
+    storage_numbers = {
+        'E164': ['+44 113 8921113', '+441138921113'],
+        'RFC3966': ['+44 113 8921113', 'tel:+44-113-892-1113'],
+        'INTERNATIONAL': ['+44 113 8921113', '+44 113 892 1113'],
+    }
```
tdobrovolny: Predefined constants for checking correct stored value

thedrow:
Thanks!

TravisDart:
I think this calls for a new release. Might I request that this be pushed to Pypi? I installed pip's "latest" version of django-phonenumber-field and it took me a while to figure out why this feature wasn't working.

thedrow:
@TravisDart unfortunately I do not have release privileges yet. I emailed @stefanfoulis about it but got no response.

thedrow:
Oh wait, I do have them. I completely forgot about this. Sorry :)
I'll release as soon as I can.

PR #83, opened 2015-09-10
Issue/field equality correction & improvements

tdobrovolny:
## Issue with phone number field equality

Override parent (phonenumbers.phonenumber.PhoneNumber) equality test because we store only string representation of phone number, so we must compare only this string representation not whole structure as parent class does. If we use parent comparison then same phone number ones created from string and ones created from python code (not by string, but by example by number & country code) are by parent comparison regarded as not equal.

### Create new validator for possible numbers

Create validator that allows any possible phone number (e.g. it has the right number of digits). This new validator is less stricter than validate_international_phonenumber.

### Improve tests

 - change test_number_1 to valid example phone number
 - add to equal_number_strings representation in rfc3966 format
 - in test_field_returns_correct_type change local string constant number to self.test_number_1
 - in test_can_assign_* add full_clean & save to db & compare & read from db & compare with entered values

### Improve from_string function

 - add assert to check that passed argument is string
 - return None if phone_number.strip() is empty string (this is consistent with to_python behavior for empty values).

Create tests for empty and strippable string.

### Simplyfy string formating functions

String formating functions __unicode__ , format_as unnecessary checks validity of number and if number wasn't valid they returns raw_input value. But same checks and functionality provides function phonenumbers.format_number which they use.


phonenumber_field/phonenumber.py
```diff
@@ -29,32 +29,38 @@ class PhoneNumber(phonenumbers.phonenumber.PhoneNumber):
...
+                               keep_raw_input=True, numobj=phone_number_obj)
         return phone_number_obj
 
     def __unicode__(self):
         format_string = getattr(settings, 'PHONENUMBER_DEFAULT_FORMAT', 'E164')
         fmt = self.format_map[format_string]
-        if self.is_valid():
-            return self.format_as(fmt)
-        return self.raw_input
+        return self.format_as(fmt)
```
tdobrovolny: Checks are done in format_as

phonenumber_field/phonenumber.py
```diff
@@ -29,32 +29,38 @@ class PhoneNumber(phonenumbers.phonenumber.PhoneNumber):
...
+        (e.g. it has the right number of digits)
+        """
+        return phonenumbers.is_possible_number(self)
+
     def format_as(self, format):
-        if self.is_valid():
-            return phonenumbers.format_number(self, format)
-        else:
-            return self.raw_input
+        return phonenumbers.format_number(self, format)
```
tdobrovolny: Checks are done in format_number function and this function also returns raw_input if number is not valid

tdobrovolny:
Needs to rework to be in line with #74.

PR #101, opened 2016-01-06
Update russian localization

Alidal:


amateja:
Fixed already in #98. Thanks.

PR #103, opened 2016-01-09
Fix #102 - Respect null field option.

### LINKED ISSUE #102: Impossible to set PhoneNumberField to None/null
tl;dr: Given the current implementation of the field, it is impossible to set the field to the database `null` value.

In the current implementation, an empty `PhoneNumberField` field will *always* be set to the empty string. While Django recommends against using subclasses of `CharField` with the option `null=True`, this sometimes becomes necessary when the field is both optional and unique.

    phone = PhoneNumberField(blank=True, default=None, null=True, unique=True)

In the case of an optional unique field, it becomes important to avoid the empty string, as this causes uniqueness `IntegrityError` to be raised. Instead, the use of `null` (or `None` in Python) become necessary, as `null` values do not cause uniqueness conflicts.

It is therefore important for the field to be able to support the ability to be set to `None` or `null`.

jambonrose:
It is now possible to set null=True on PhoneNumberField instances.

### DIFF
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -3,10 +3,12 @@
 from django.conf import settings
 from django.core import validators
 from django.db import models
+from django.db.models.fields import NOT_PROVIDED
 from django.utils.translation import ugettext_lazy as _
-from phonenumber_field.validators import validate_international_phonenumber
+
 from phonenumber_field import formfields
-from phonenumber_field.phonenumber import PhoneNumber, to_python, string_types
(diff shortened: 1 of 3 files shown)


jambonrose:
The tests in this PR depend on #105 , and will not work until is is merged (or #104 is fixed by some other means).

jambonrose:
I'm using 3 helper functions in the testing suite to de-duplicate code: I can move this code into the actual tests if you want tests on Quantified Code to pass.

The other (13) errors on Quantified Code are for empty string 'constants' used in the `default` options of test models.

PR #107, opened 2016-01-31
HTML5-style widget ('tel' instead of 'text')

interDist:


stefanfoulis:
How do browsers that don't support the ``tel`` input type react?

interDist:
Older browsers (with no support for HTML5 input elements) fallback to the "text" input type. Attributes that are not supported (e.g., ‘placeholder’) are simply discarded.

See http://caniuse.com/#search=type%3Dtel (section ‘notes’) and https://mathiasbynens.be/notes/html5-levels#level-2 (“Level 2: Stuff that degrades gracefully by default”).

I tested this myself with a set of browsers >8 years old.

stefanfoulis:
cool. thanks for the clarification @interDist

PR #110, opened 2016-02-16
Allow empty value for PhoneNumberPrefixWidget

davitv:
When you are using PhoneNumberPrefixWidget you are not able to leave phone field blank because of dot added between phone prefix and actual phone value


amateja:

PR #117, opened 2016-03-24
Add a new widget for smart switching between national and international formats

MatthewWilkes:
Currently, if using `PHONENUMBER_DEFAULT_FORMAT = 'NATIONAL'` numbers that are entered in international format lose their country information on loading the form, meaning they cannot be validated. 
This widget checks if the number provided is in the expected region and renders it in national format if so, or international if there's a mismatch.

### DIFF
--- a/phonenumber_field/widgets.py
+++ b/phonenumber_field/widgets.py
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
 
 from babel import Locale
+from django.conf import settings
 from django.forms import Select, TextInput
 from django.forms.widgets import MultiWidget
 from django.utils import translation
 from phonenumbers.data import _COUNTRY_CODE_TO_REGION_CODE
+from phonenumbers import PhoneNumberFormat
+from phonenumbers.phonenumberutil import region_code_for_number
 
 from phonenumber_field.phonenumber import PhoneNumber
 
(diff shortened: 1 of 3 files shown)


amateja:
Prefect. Thank you. Could you also add some test for your new code?

MatthewWilkes:
@amateja I already included a test for the widget, in fact that's why the coverage has dropped so much, previously there was no reference to widgets in the tests so the widgets.py was excluded from the totals.

If there are some specific additional cases you'd like me to add I'd be happy to!

amateja:
All right then. I'm sorry haven't had time to review this PR properly but I promise I will.

MatthewWilkes:
@amateja No worries, tbh I was shocked that there'd been an acknowledgement so quickly. I'm happy to run the widget in the project code I need it for, so I have no rush. I may end up sending more, depending on if I hit other edge cases.

phonenumber_field/widgets.py
```diff
@@ -57,3 +60,27 @@ def value_from_datadict(self, data, files, name):
...
+class PhoneNumberInternationalFallbackWidget(TextInput):
+    """
+    A Widget that allows a phone number in a national format, but if given
+    an international number will fall back to international format
+    """
+
+    def __init__(self, region=None, attrs=None):
+        if region is None:
+            region = (getattr(settings, 'PHONENUMBER_DEFAULT_REGION', None)
+                      or getattr(settings, 'PHONENUMER_DEFAULT_REGION', None))
```
amateja: @MatthewWilkes are you sure that this is correct?

amateja: Maybe I'm wrong but I think that part when you assign value to region variable is duplicated.

MatthewWilkes:
I thought that was weird too, but it wasn't my idea. It's copied from https://github.com/stefanfoulis/django-phonenumber-field/blob/develop/phonenumber_field/phonenumber.py#L34,L35

I guess it's to maintain compatibility with a typoed setting fom way back when.

I'm happy to lose that compatibility, especially as it's new code.

kevin-brown:
> I guess it's to maintain compatibility with a typoed setting fom way back when.

Specifically, it was added in https://github.com/stefanfoulis/django-phonenumber-field/pull/25.

amateja:
Now I see why it is that way. Thank you. I propose to break this compatibility here and everywhere else. What do you think?

MatthewWilkes:
@amateja Personally, I think that's a sensible idea. My feeling is that not too many people will be doing blind upgrades.

I'm happy to submit a PR tomorrow to split that conversation out. I will also update this one to remove the backwards compatibility, as at no time was this widget available when the typo'ed setting name was recommended.

Issue #123, opened 2016-04-22
Running migration on phonenumberfield results in AttributeError


dedayoa:
I've got the straight up install of django-phonenumber-field, with phonenumberslite.
See extract from running pipdeptree below.

`django-phonenumber-field==1.1.0
  - phonenumberslite [required: >=7.0.2, installed: 7.3.1]
  - babel [installed: 2.3.3]
    - pytz [required: >=0a, installed: 2016.3]`

When I run manage.py migrate, I get 

`  File "C:\Users\...\lib\site-packages\phonenumber_field\modelfields.py", line 76, in get_prep_value
    return value.format_as(fmt)
AttributeError: 'NoneType' object has no attribute 'format_as'`

as the last entry in the error trace.

I'm running Python 3.4.3
Any suggestion to what is causing this? How can I remedy as it is preventing my migrations from successfully running.

banagale:
Thank you for the report and fix on this.

PR #124, opened 2016-04-22
Update modelfields.py to address Issue #123

dedayoa:
The patch to get_prep_value() allows manage.py migrate to successfully complete. This is to address issue "Running migration on phonenumberfield results in AttributeError #123"

As per https://docs.djangoproject.com/en/1.8/howto/custom-model-fields/#custom-field-deconstruct-method, I also added the deconstruct method. I don't know if my migration will run without this, but figure since it's required, it should stay.

### DIFF
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -73,6 +73,10 @@ class PhoneNumberField(models.Field):
             return value
         format_string = getattr(settings, 'PHONENUMBER_DB_FORMAT', 'E164')
         fmt = PhoneNumber.format_map[format_string]
+        
+        if value is None:
+            return ''
+        
         return value.format_as(fmt)
 
     def contribute_to_class(self, cls, name):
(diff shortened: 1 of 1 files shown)


phonenumber_field/modelfields.py
```diff
@@ -73,6 +73,10 @@ def get_prep_value(self, value):
             return value
         format_string = getattr(settings, 'PHONENUMBER_DB_FORMAT', 'E164')
         fmt = PhoneNumber.format_map[format_string]
+        
+        if value is None:
+            return ''
```
amateja: What if a field has `null=True?` Shouldn't it return `None` then?

phonenumber_field/modelfields.py
```diff
@@ -85,6 +89,11 @@ def formfield(self, **kwargs):
         }
         defaults.update(kwargs)
         return super(PhoneNumberField, self).formfield(**defaults)
+        
+    def deconstruct(self):
```
amateja: A `deconstruct` method is required from django 1.7 for all custom fields but PhoneNumberField is based on very common `CharField`. Moreover this particular `deconstruct` method does nothing... well it removes `max_length` parameter which is wrong. PhoneNumberField uses 128 by default but if you want you can set it to different value. Deleting `max_length` parameter you are loosing track of what user's defined.

amateja:
Hi @dedayoa,

I cannot refer to your problem not having a piece of your code. I've already commented your changes and I'm not very convinced to your PR. Are you sure that it is not a bug in your application? Recently we have changed `get_prep_value` method to respect `null=True` option in #102 and it appears to me like an aftermath.

Issue #71, opened 2015-05-04
Issue with using PhoneNumberPrefixWidget in a formset


lchan752:
Thanks for the great work. Wonderful phone number field.

I was using this in a formset. And if the formset contains empty forms, django's default behavior is to short circuit any validation.

See https://github.com/django/django/blob/master/django/forms/forms.py#L373

But has_changed() is always returning true when using the PhoneNumberPrefixWidget, even if no values is submitted in the form. 

I think the bug is here.
```python
def value_from_datadict(self, data, files, name):
    values = super(PhoneNumberPrefixWidget, self).value_from_datadict(data, files, name)
    return '%s.%s' % tuple(values)
```
This code will return "." if values is None, so has_changed() is comparing "." with the initial value of the field, which is None, and is returning false because "." != None

I think the fix should be this.
```python
def value_from_datadict(self, data, files, name):
    values = super(PhoneNumberPrefixWidget, self).value_from_datadict(data, files, name)
    if not any(values):
        return None
    return '%s.%s' % tuple(values)
```

amateja:
Thank you for reporting. Our project has evolved since that time and this issue will be closed. Please update and reopen if you feel it is still applicable. You may also create PR instead :smile: .

PR #126, opened 2016-05-12
Correct form validation with empty fields

walterrenner:
This PR refers to #71.

Empty forms of an formset were treated as not valid because the sting "." was retuned and not None.


stefanfoulis:
This may be fixed by #140 . Please re-submit if the issue still exists.

PR #161, opened 2017-02-27
Fix #159: rewrite PhoneNumberField.get_prep_value, add tests

### LINKED ISSUE #159: How to filter records with empty value for PhoneNumberField?
Model.objects.filter(phone="") does not work.
Model.objects.filter(phone__isnull = True) neither.

How to do this?

lampslave:


lampslave:
BTW, you have huge amount of tests on Travis CI, but all of them use SQLite. Maybe it's time to use actual versions of Django and actual databases?

PR #168, opened 2017-04-10
test against django 1.11

thijstriemstra:


thijstriemstra:
Failing with https://travis-ci.org/stefanfoulis/django-phonenumber-field/jobs/220683434:
```
  File "<frozen importlib._bootstrap>", line 2254, in _gcd_import

  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load

  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked

  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked

  File "<frozen importlib._bootstrap>", line 1129, in _exec

  File "<frozen importlib._bootstrap>", line 1471, in exec_module

  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed

  File "/home/travis/build/stefanfoulis/django-phonenumber-field/testproject/testproject/urls.py", line 1, in <module>

    from django.conf.urls import patterns, include, url

ImportError: cannot import name 'patterns'
```

Should support for older versions be dropped or should it work on all versions of Django?

amateja:
Import error is not a very big deal. If it's the only compatibility issue we can work it around. Could you test it on your own and give us a full list of possibly deprecated django version if there is more problems?

thijstriemstra:
> Import error is not a very big deal. 

Huh? It prevents the tests from running.

amateja:
I mean it could be easily avoided. Starting from Django 1.8 defining `urlpatterns` as a list not a result of `patterns` function is preferred. So it's nothing new. The easiest workaround I can think of is to wrap import of it in `try` - `catch` clause and define dummy `patterns` function in else clause. This is just test app not a core functionality so such hack should be forgiven 😄.

thijstriemstra:
Yeah, so you want to keep compat for 1.8? Ill update the PR.

amateja:
It's not a matter what I want. Django 1.8 is still [supported](https://www.djangoproject.com/download/#supported-versions) so this project should comply. We might drop all obsolete versions if maintaining them costs us much effort... which is not the case now.

thijstriemstra:
> It's not a matter of what I want. 

I get a really weird vibe here. Feel free to close the PR.

tredzko:
@thijstriemstra I think @amateja is getting at is that we can swap how we populate url patterns based on which version of django the test is running since the PR is failing because of the test implementation, not the package itself. This doesn't affect the package itself. If, once we make sure the tests can be run, the core library has issues, then we'd worry about compatibility and make necessary changes.

stefanfoulis:
@thijstriemstra thanks for adding 1.11 to the test matrix. I made ``urlpatterns`` conditional based on the django version as @amateja suggested.

Unfortunately the tests did not pass with Django 1.11 so I marked them as allowed failures for now.
The first failing test just looks like a different order of html attributes though. So the tests will need to be adapted to be a bit more fuzzy about the order of html attributes.

```
FAIL: test_fallback_widget_switches_between_national_and_international (phonenumber_field.tests.PhoneNumberFieldTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/stefanfoulis/django-phonenumber-field/phonenumber_field/tests.py", line 175, in test_fallback_widget_switches_between_national_and_international
    u'<input name="number" type="text" value="01606 75178" />'
AssertionError: u'<input type="text" name="number" value="01606 75178" />' != u'<input name="number" type="text" value="01606 75178" />'
```

Regarding supported versions: I'm fine with dropping Django <1.8 support soonish.  And with limiting python support to 2.7 and 3.5.

PR #172, opened 2017-04-27
Don't raise an AttributeError when accessing a PhoneNumberDescriptor

kmichel-sereema:
This fixes compatibility with Django 1.11 when initializing a model
instance with kwargs values such as TestModel(name='name').

Model initialization was changed in Django 1.11 and now tries to
access fields to check their class in Options._property_names()
in django/db/model/options.py.

https://github.com/django/django/blob/master/django/db/models/options.py#L837


smn:
Anything blocking this getting merged in?

PR #188, opened 2017-08-21
Recompile translations

gfavre:
.mo files haven't been recompiled for every language. As a result for instance the french version of "The phone number entered is not valid." was not available.


thedrow:
Thanks!

Issue #171, opened 2017-04-21
German localization files


deluge:
The *.po file for de is marked as fuzzy. So If I use de locale I get english error messages etc.

amateja:
Thanks for your report. I know where problem might be. Could you double check our de localization file to verify if translation is made properly? https://github.com/stefanfoulis/django-phonenumber-field/blob/master/phonenumber_field/locale/de/LC_MESSAGES/django.po

Mactory:
I just checked the translation file you posted. The German translation is correct. You can remove the fuzzy mark.

PR #189, opened 2017-10-05
Fixing German translation files

Mactory:
Re-ran `django-admin makemessages` and  removed the `fuzzy` marker


Mactory:
Fixes #171

thedrow:
Thanks!

PR #194, opened 2017-11-30
Add new PhoneNumberRegionFallbackField

jonashaag:
This adds a new field that replaces `PhoneNumberInternationalFallbackWidget` by providing a superset of its features.

## My use case

- Render phone numbers without region code prefix if the user is in the region (e.g. render +49 12345 as 12345 if the user is from Germany)
- Allow input without region code prefix and default to the user's region code (e.g. an input of "12345" is interpreted as "+49 12345" in Germany and "+44 12345" in the UK)
- Allow input with region code prefix, in which case the user's region code is ignored (e.g. an input of "+49 12345" is interpreted as "+49 12345" for all users)

## Why a new field?

- `PhoneNumberInternationalFallbackWidget` only provides a subset of these features (the first point, region-dependent rendering)
- From my understanding of the problem `PhoneNumberInternationalFallbackWidget` tries to solve it's the wrong place (read: abstraction) to solve that kind of problem. From the Django docs: "A widget is Django’s representation of an HTML input element. The widget handles the rendering of the HTML, and the extraction of data from a GET/POST dictionary that corresponds to the widget." If you look at the built-in widgets Django provides, they mostly deal with HTML-level plumbing like `input type=xxx` attributes, checkbox `on` parsing, etc. NOT with the question of what text/value should be displayed to the user, and especially not parsing/interpreting values put into input fields by users. That's the job of `Field` types.
- I think we should drop `PhoneNumberInternationalFallbackWidget` in favour of this replacement.


jonashaag:
@stefanfoulis Hooray, no more test failures (except for an unrelated PyPy 3 failure)

jonashaag:
bump

jonashaag:
@stefanfoulis bump

stefanfoulis:
@jonashaag sorry I missed this PR so far. I agree with avoiding non-display logic in the widget and am fine with deprecating the other widget.
In terms of compatibility I'd also be fine with dropping support for Django<1.8 or even <1.11

francoisfreitag:
Probably solved by the new `RegionalPhoneNumberWidget`. If something’s missing, feel free to open an issue. This PR is quite outdated now.

PR #196, opened 2017-12-04
Add support for Django 2.0.

### LINKED ISSUE #193: Django 2.0 support: _format_value renamed, maybe more
Django 2.0 is slated for release in about a month and django-phonenumber-field should work with it.

As far as I know, django-phonenumber-field uses at least one feature that had a breaking change in Django 2.0: [Widget.format_value](https://docs.djangoproject.com/en/1.11/ref/forms/widgets/#django.forms.Widget.format_value) has been renamed from `_format_value` to `format_value` to make it a part of the public API.

The CI test matrix is growing quite large, and with the addition of Python 3.6 and Django 2.0 it will become even larger. Django upstream suggests to drop support for anything older than 1.11 LTS and make sure the project runs on 1.11 without any warnings. Then it should work on 2.x as well without changes. I'll leave to the project maintainers to decide if the list of Django versions supported by django-phonenumber-field should be pruned a bit.

benkonrath:
Support for versions of Django < 1.11 have been dropped as recommended.

Closes #193


benkonrath:
It should be possible to keep support for the older version of Django if you want that. Let me know if this is the case and I'll update the PR. Thanks.

PR #209, opened 2018-04-17
set region into formfield

satels:


francoisfreitag:
Hi @satels,

I believe https://github.com/stefanfoulis/django-phonenumber-field/pull/263 may solve your use case?

sandeepimpressads:
Is this feature going to be rolled out ?

PR #243, opened 2018-12-17
Fix value when null is True

lucifurtun:


stefanfoulis:
Thanks @lucifurtun :-)

Looks like the tests for Django 1.11 are failing. Weird 🤔 

(pypy3 is passing, but that is because of #244)

lucifurtun:
@stefanfoulis, ok, but what do you think about this approach ? I'll check the tests if there might be any chance to merge this PR.

stefanfoulis:
Hmm, now that you mention it, I'm not sure.

Currently, the assumption is that you don't set `null=True` on `PhoneNumberField` because the `CharField` (which is used under the hood) can perfectly well save the "empty" state with an empty string. So `null` is an unnecessary extra state which adds complexity.

@lucifurtun What is your use case for defining the field with `null=True`?


(I can think of one reason: When initially adding it as a new field on a table a lot of rows it is more efficient in most db backends to create the new column as nullable.)

lucifurtun:
@stefanfoulis, Well, yes, that's kind of it. We want to have `NULL` value in database when there is no phone number. Also, this is the django admin default behaviour, when a `CharField` has `null=True`, if there is no value, in db the value will be `NULL`. I think `PhoneNumberField` should be consistent with `CharField`.

stefanfoulis:
@lucifurtun Yes, I agree. We should be consistent with `CharField` behaviour.
👍

nicolasmendoza:
Also it doesn't support blank values. Why ? @stefanfoulis

lucifurtun:
@stefanfoulis, I have a better approach. The value shouldn't be enforced on the model field level. It's better to handle this in the form. This is similar to what `CharField` does. What do you think ? I also added a test to prove it.

stefanfoulis:
Lookin good :-)

Do you see any possible regressions this change could have for existing users?

lucifurtun:
I don't see how this change can affect existing users. Only if they rely on having empty string in db when there is no phone number.

Also, I'm not sure I'll have time to invest in more tests, I'll see what I can do.

stefanfoulis:
@lucifurtun Additional test: no problem. I think this PR is good to merge as is.

lucifurtun:
@stefanfoulis , does anything prevent from merging and creating a new release ?

stefanfoulis:
@lucifurtun I think we can release. I try to find the time to do so over the next few days.

lucifurtun:
Ok, thanks!

Issue #244, opened 2018-12-18
Fix test setup for pypy3 (or remove it from the matrix)


stefanfoulis:
It looks like pypy3 is not actually running the tests on travis:

```
$ tox -- --cov
___________________________________ summary ____________________________________
  congratulations :)
The command "tox -- --cov" exited with 0.
```

We should either fix this or remove pypy3 from the matrix.
#104 mentions that babel is not supported on pypy, which may be the problem.

PR #245, opened 2018-12-21
Get Python 3.7 officially running on Travis CI

kevin-brown:
- [x] Switch to always running on Xenial
- [x] Add Django 1.11 + Python 3.7 to the matrix


stefanfoulis:
pypy3 is still not actually running any tests. But that is a different issue: #244

Issue #133, opened 2016-06-23
PhoneNumberPrefixWidget sometimes selects wrong country as initial value


benmurden:
This happens with, for example, United Kingdom, which has a country code of +44, but so does Guernsey, meaning the select box defaults to that when returning to the form.

tisdall:
This keeps coming up:  #125 #82

Basically the field won't work properly because it only stores the country code and not the country.  When the field is re-rendered, it selects the first item in the list alphabetically that matches that country code.

Leave this issue open so the next person to discover this will see that it's already a known issue.

benmurden:
Thanks for the response and sorry for the repeat issue.

I suspect the only way around it would be to have another, unique identifier for each country code, rather than rely on the code itself, which is clearly not unique.

tisdall:
The problem, though, is the database model only stores a phone number and not a country code.

I think the `phonenumbers` package has methods to figure out a country based on just the phone number, but it's rather intensive on memory and has to sort through XML files.

If you really want this type of field, it'd probably be best to look at the source code here and make your own similar one that had 2-letter country iso codes instead of international dialing codes and then compile the phone number yourself with `phonenumbers` and store the number along with the country code in the database.

benmurden:
I created the following widget to resolve this issue for us.

https://github.com/benmurden/django-intl-tel-input

It uses the intl-tel-input jQuery plugin to output the widget, though we're still using django-phonenumber-field for backend validation.

thenewguy:
I haven't looked at this in a quite some time so I am not familiar with it at the moment, but in my fork I modified the field to store the country in the database with the phone number if available.  I also made a widget that makes use of this to display the appropriate country in a drop down.  I don't remember this being very involved if someone wanted to take a look.  https://github.com/thenewguy/django-phonenumber-field

Another approach that might be more compatible with the current code base (I'm no longer familiar with it so just an idea at best) would be to store the country code in a separate field that the phone number field has access to... sort of like django image field and width/height data.

PR #248, opened 2019-01-10
Add by region lookup method for phonenumberfields

cmehay:
Phone numbers can be found using associated region like 'FR', 'UK', 'US', etc.

	queryset.filter(phone_number_field__region='FR)
	queryset.get(phone_numer_field__region='UK')

Using startswith pattern with region codes

	LIKE '+44%';

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,7 +43,7 @@ As a an alternative to the ``phonenumbers`` package it is possible to install th
 Basic usage
 ===========
 
-First, add ``phonenumber_field`` to the list of the installed apps in 
+First, add ``phonenumber_field`` to the list of the installed apps in
 your ``settings.py`` file::
 
     INSTALLED_APPS = [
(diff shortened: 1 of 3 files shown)


cmehay:
I implemented this because I had issues with `__startswith` lookup method with phonenumber_fields to find numbers by regions.

amateja:
I really like this idea. :+1:

Could you also consider a scenario where in the same database you have numbers from US and Canada (both have +1 prefix) or Italy and Vatican (both have +39 prefix) or Russia and Kazakhstan (both have +7 prefix) or US and Trinidad and Tobago (respectively +1 and +1868) or from Dominican Republic having more than one country code (+1809, +1829, +1849)? How about France (+33) and French overseas departments (e.g. Saint-Martin +590)?

cmehay:
You're right, I didn't think about those cases. I'll need to find another way.

stefanfoulis:
We've had similar problems with the fact that phone prefixes don't match 1:1 with regions with the `PhoneNumberPrefixWidget` (#133). It would be really nice if we found a solution for it.

cmehay:
I updated the method to use regex lookup to find numbers by region. But sadly the tests don't pass yet, the sql request seems good but sqlite returns more than expected results in tests.

cmehay:
I added tests for countries with same code like canada and usa and sadly it doesn't work 😞

stefanfoulis:
@cmehay yeah, but that was to be expected. Since we don't actually save the region, only the full phone number. I think this is still useful though. We should probably mention this caveat in the README though.

I think the long term solution will be adding a new field that additionally can save the chosen region to the database. When we do that we can make this region lookup work more precisely if the data is there.

francoisfreitag:
Closed for inactivity.

PR #263, opened 2019-02-18
Add region argument to PhoneNumberField

francoisfreitag:
The `PHONENUMBER_DEFAULT_REGION` setting is applied system-wide, when the application is loaded. In my use case, the default region is set by users and may change during runtime.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 CHANGELOG
 =========
 
+* Added the ``region`` keyword argument to ``PhoneNumberField``.
+
 2.2.0 (2019-01-27)
 ------------------
 
--- a/phonenumber_field/formfields.py
+++ b/phonenumber_field/formfields.py
@@ -6,7 +6,7 @@ from django.core.exceptions import ValidationError
 from django.forms.fields import CharField
 from django.utils.translation import ugettext_lazy as _
 
-from phonenumber_field.phonenumber import to_python
+from phonenumber_field.phonenumber import to_python, validate_region
 from phonenumber_field.validators import validate_international_phonenumber
 
 
@@ -15,11 +15,13 @@ class PhoneNumberField(CharField):
     default_validators = [validate_international_phonenumber]
 
     def __init__(self, *args, **kwargs):
+        self.region = kwargs.pop("region", None)
+        validate_region(self.region)
         super(PhoneNumberField, self).__init__(*args, **kwargs)
         self.widget.input_type = "tel"
 
     def to_python(self, value):
-        phone_number = to_python(value)
+        phone_number = to_python(value, region=self.region)
 
         if phone_number in validators.EMPTY_VALUES:
             return self.empty_value
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -2,12 +2,13 @@
 from __future__ import unicode_literals
 
 from django.conf import settings
-from django.core import validators
+from django.core import checks, validators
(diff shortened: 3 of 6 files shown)


stefanfoulis:
Will this cause a django database migration for people already using PhoneNumberField in their projects (because of the changed signature)?

francoisfreitag:
Currently, Django won't see any difference. `region` [is not checked by Django when deconstructing the field](https://github.com/django/django/blob/14e2b1b065085c1d2d3e94ebaeefe25e12595a00/django/db/models/fields/__init__.py#L432-L453). The [documentation states](https://docs.djangoproject.com/en/dev/topics/migrations/)

> Django will make migrations for any change to your models or fields - even options that don’t affect the database - as the only way it can reconstruct a field correctly is to have all the changes in the history, and you might need those options in some data migrations later on (for example, if you’ve set custom validators).

That gets me thinking a custom `deconstruct` method should be added for `region`, to avoid changing the interpretation of a national phone number stored in the database if the field's `region` changes. What do you think?

francoisfreitag:
Added a custom `deconstruct` method for the field and validated region code against available regions in `python-phonenumbers`.

Not specifying the `region`, or leaving it as `None` won't generate migrations for existing projects:

```bash
$ git checkout master && git clean -dfx
$ cd testproject
# Edit testproject settings to use a database, by setting a database name.
$ ./manage.py makemigrations testapp
Migrations for 'testapp':
  testapp/migrations/0001_initial.py
    - Create model TestModel
    - Create model TestModelPhoneB
    - Create model TestModelPhoneBDNU
    - Create model TestModelPhoneBEDNU
    - Create model TestModelPhoneBNDNU
    - Create model TestModelPhoneBNU
    - Create model TestModelPhoneDNU
    - Create model TestModelPhoneEDNU
    - Create model TestModelPhoneNDNU
    - Create model TestModelPhoneNU
$ ./manage.py migrate
Operations to perform:
  Apply all migrations: auth, contenttypes, sessions, sites, testapp
Running migrations:
  Applying contenttypes.0001_initial… OK
  Applying contenttypes.0002_remove_content_type_name… OK
  Applying auth.0001_initial… OK
  Applying auth.0002_alter_permission_name_max_length… OK
  Applying auth.0003_alter_user_email_max_length… OK
  Applying auth.0004_alter_user_username_opts… OK
  Applying auth.0005_alter_user_last_login_null… OK
  Applying auth.0006_require_contenttypes_0002… OK
  Applying auth.0007_alter_validators_add_error_messages… OK
  Applying auth.0008_alter_user_username_max_length… OK
  Applying auth.0009_alter_user_last_name_max_length… OK
  Applying auth.0010_alter_group_name_max_length… OK
  Applying auth.0011_update_proxy_permissions… OK
  Applying sessions.0001_initial… OK
  Applying sites.0001_initial… OK
  Applying sites.0002_alter_domain_unique… OK
  Applying testapp.0001_initial… OK
# Switch to this branch
$ git checkout region_form_field
M       testproject/testproject/settings.py
Switched to branch 'region_form_field'
$ ./manage.py makemigrations testapp
Migrations for 'testapp':
  testapp/migrations/0002_frenchphoneowner.py
    - Create model FrenchPhoneOwner
$ ./manage.py migrate
Operations to perform:
  Apply all migrations: auth, contenttypes, sessions, sites, testapp
Running migrations:
  Applying testapp.0002_frenchphoneowner… OK

# Existing models do not generate a new migration.
# Changing FrenchPhoneOwner region from FR to CH triggers a new migration.
```

PR #264, opened 2019-02-18
Remove model field to_python call in get_prep_value

francoisfreitag:
The model instance has been set through the descriptor's `__set__`, `to_python` has already been called there, I believe this call is a duplicate.


francoisfreitag:
Outdated per https://github.com/stefanfoulis/django-phonenumber-field/pull/273.

PR #266, opened 2019-02-19
Remove INSTALLED_APPS section from README

francoisfreitag:
This app has no models and is not configurable through an `AppConfig`, there's no benefit to add it to Django's `INSTALLED_APPS`. Removing that section makes installation quicker.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 CHANGELOG
 =========
 
+* Remove recommendation to add ``phonenumber_field`` to the ``INSTALLED_APPS``.
+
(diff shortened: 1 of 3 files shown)


jdufresne:
I believe the Travis failures will be resolved by #257 :slightly_smiling_face:

stefanfoulis:
I think there was a reason to put it in `INSTALLED_APPS`.... 🤔

stefanfoulis:
Ah.. I think so that django picks up the `locale` folder for the translations.

francoisfreitag:
Thank you for explaining!

PR #267, opened 2019-02-20
Use dict setdefault instead of reimplementing it

francoisfreitag:
It's clearer in intent, and slightly faster.

```bash
$ python -m timeit 'd = {}; d.setdefault("a", 12)'
5000000 loops, best of 5: 96.2 nsec per loop
$ python -m timeit 'd = {}; d["a"] = d.get("a", 12)'
2000000 loops, best of 5: 107 nsec per loop
```


stefanfoulis:
makes sense. thanks!

PR #268, opened 2019-02-20
Inherit from Django CharField

francoisfreitag:
Simplifies initialization, avoids copying some CharField methods over and let Django handle database compatibility regarding storage of null values.

### DIFF
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -2,8 +2,7 @@
 from __future__ import unicode_literals
 
 from django.conf import settings
-from django.core import validators
-from django.db import connection, models
+from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
 from phonenumber_field import formfields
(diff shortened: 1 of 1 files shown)


francoisfreitag:
The travis-ci failure was due to a network error when fetching pypy35, nothing related with the code in this PR. I force-pushed to trigger another build.

The coverage drop is not significant. I removed code from the package, which reduced the number of lines covered. Please let me know if there are any concerns in that regard.

stefanfoulis:
Weird that the coverage went down... You removed code, so it should have stayed the same (or gone up). Anyway, I don't see a problem.

Have you tested if this change will cause django to want to add a database migration for the models that are using the field?

francoisfreitag:
The coverage went down because it is computed as the number of covered lines / number of total lines. Since the removed lines were covered, the proportion of covered lines was reduced.

No migrations should be generated for this change, nothing changes from the DB point of view, still a charfield, same maxlength, same nullable properties, default, and I don't see a reason for Django to generate a migration... The following experiment seems to confirm:
```bash
$ git checkout master
$ cd testproject/
# Configure testproject/settings.py to use a database, either specifying a name with sqlite or another database engine
$ ./manage.py makemigrations testapp
# ... creates initial migration for testapp models
$ ./manage.py migrate
$ git checkout inherit_charfield
$ ./manage.py makemigrations testapp
# No changes detected in app 'testapp'
```

jdufresne:
Rebased and merged as fe946db5bb59740150cd7e1b05c1e39ee7f291bb. Thanks!

PR #273, opened 2019-02-21
Only save valid phone numbers to the database

francoisfreitag:
Saving an invalid phone number resulted in the value `+NoneNone` being saved to the database. This is an invalid record.

~Assume that value is either empty (`None` or `""`) or a `PhoneNumber`. The field is always accessed through its descriptor, which always return a `PhoneNumber` when the value's not empty.~
The field can be used in a [`QuerySet.update()`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#update) statement.

Avoids extra work of coercing the value to a `PhoneNumber`.


francoisfreitag:
Updated:

- `value` can be an `str` when `QuerySet.update` is used. Updated code to coerce to Python when getting an `str` value and added tests for the different scenarios (`update()`, `save(update_field=["phone"])`).

jdufresne:
Rebased as 693b9e5c137810653ef1ed915d3c22def59d0d98. Thanks!

PR #309, opened 2019-05-28
Drop support for Python 3.4

francoisfreitag:
Python 3.4 is not maintained anymore.
https://www.python.org/downloads/release/python-3410/


jdufresne:
Thanks, but this work has already been done in #307.

francoisfreitag:
I only read the PR title :confounded:.

PR #310, opened 2019-05-28
Drop support for Django 2.0

francoisfreitag:
Django 2.0 is not maintained since Apr 1st, 2019.
https://www.djangoproject.com/download/#supported-versions


jdufresne:
Thanks

PR #311, opened 2019-05-28
 Give example in invalid phone number error message 

francoisfreitag:
Helps user understand why their input does not pass validation and the expected format.


jdufresne:
Thank you! Rebased and merged as cb35a6a41e7e91ee3b4ec863fc12f5d533eff7cd

Issue #202, opened 2018-02-19
initialize PhoneNumberPrefixWidget with initial don't make effect


boy2000-007man:
Hi,
Thanks for the work.:)

When setting the default region number with parameter initial, I find something strange.
My code is shown below:
```
phone = PhoneNumberField(
    label=_('phone number'),
    required=True,
    widget=PhoneNumberPrefixWidget(
        initial='CN',
        attrs={'placeholder': _('Phone number')}))
```
The default prefix is still empty.:(
I trace the code and find PhonePrefixSelect.render() won't be accessed during runtime, which should be to pass the default prefix.
Later I find the default value of the prefix may be fixed by changing the PhoneNumberPrefixWidget.decompress()'s return value.
I don't understand what happens, does latest Django v2.0.2 change the rendering logic so that the PhonePrefixSelect.render() loses its function?
Or there is something wrong with my code I don't notice?

Thank You

quevon24:
Worked form me this way:

cellphone = PhoneNumberField(
        widget=PhoneNumberPrefixWidget(attrs={'placeholder': _(u'Cellphone'), 'class': "form-control"}),
        label=_(u'Cellphone number'),
        required=False,
        initial='+52'
    )

boy2000-007man:
Thanks a lot. I'll try it later. Your solution seems using the CharFields's initial keyword. And I think the PhoneNumberPrefixWidget's own initial keyword may not correctly be implemented.

QFXC:
This is still an issue. I am using Django version 2.0.9. I had to come up with this hacky way of getting a default/initial value.

`
# forms.py
from .widgets import CustomPhoneNumberPrefixWidget
class AccountProfilePhoneForm(AccountModelForm):
    class Meta:
        model = AccountProfile
        fields = AccountProfile.PHONE_FORM_MINIMUM
        widgets = {'phone_num': CustomPhoneNumberPrefixWidget(initial=('+1', 'United States +1'))}

# widgets.py
from phonenumber_field.widgets import (
    PhonePrefixSelect, Select, translation, settings, 
    _COUNTRY_CODE_TO_REGION_CODE, Locale, TextInput, PhoneNumber, MultiWidget,
)

class CustomPhonePrefixSelect(PhonePrefixSelect):
    initial = None

    def __init__(self, initial=None):
        choices = [("", "---------")]
        language = translation.get_language() or settings.LANGUAGE_CODE
        if language:
[...]

PR #324, opened 2019-08-13
Correct way to set initial value for PhonePrefixSelect using "selected" attribute.

ComradeDOS:
Subj. This should help with issue #202


jdufresne:
Can you add a test, please to demonstrate the fix?

phonenumber_field/widgets.py
```diff
@@ -28,8 +30,16 @@ def __init__(self, initial=None):
...
         super().__init__(choices=sorted(choices, key=lambda item: item[1]))
 
-    def render(self, name, value, *args, **kwargs):
-        return super().render(name, value or self.initial, *args, **kwargs)
+    def create_option(self, name, value, label, selected, index, subindex=None, attrs=None):
+        if not value:
+            selected = False
+        if not self.has_selected_value and self.initial is not None and value == self.initial:
+            selected = True
+        return super(PhonePrefixSelect, self).create_option(name, value, label, selected, index, subindex, attrs)
```
jdufresne: This project is Python 3 only, so you can simplify this to:

```suggestion
        return super().create_option(name, value, label, selected, index, subindex, attrs)
```

phonenumber_field/widgets.py
```diff
@@ -28,8 +30,16 @@ def __init__(self, initial=None):
...
+
+    def optgroups(self, name, value, attrs=None):
+        self.has_selected_value = any(value)
+        return super(PhonePrefixSelect, self).optgroups(name, value, attrs)
```
jdufresne: Same here.

```suggestion
        return super().optgroups(name, value, attrs)
```

jdufresne:
This was fixed in 29b4866c06cfce3edafd813365d8f32c28efe499. If you're still experiencing issues, please report back.

PR #327, opened 2019-08-22
Format example numbers according to settings

maximdeclercq:


jdufresne:
Could you add a test?

Issue #306, opened 2019-05-22
Should raise ValidationError instead of ValueError(?)


ExTexan:
I'm wondering if the exception being raised in modelfields.py, line 86 should be a ValidationError instead of a ValueError?

I'm using the Django Rest Framework.  The reason I'm questioning this exception is because of what I see for other errors in the API Browser.  For example, this is what's returned when I leave a phone number field blank...

{
    "phone_number": ["This field may not be blank."],
    "status_code": 400
}

The above is what we want/need.  But when I enter an invalid phone number, like a North American phone number with too few digits, I get a full-page exception error, complete with Traceback, like this...

## ValueError
“+1512345678” is not a valid phone number.
/usr/local/lib/python3.6/site-packages/phonenumber_field/modelfields.py in get_prep_value, line 86
...(the rest omitted for brevity)...

jdufresne:
Can you provide the full trace please? I think this should probably be fixed in the serializer field. PR to make this change is welcome.

ExTexan:
Sure, here it is...

Request Method: PUT
Request URL: http://virticl.api/users/1/

Django Version: 2.1.5
Python Version: 3.6.8
Installed Applications:
['tenants.apps.TenantsConfig',
 'users.apps.UsersConfig',
 'api_browser.apps.APIBrowserConfig',
 'countries.apps.CountriesConfig',
 'notifications.apps.NotificationsConfig',
 'utilities.apps.UtilitiesConfig',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django_filters',
 'rest_framework',
 'rest_framework.authtoken',
 'channels',
 'debug_toolbar',
 'phonenumber_field',
 'clients.apps.ClientsConfig',
 'projects.apps.ProjectsConfig',
 'tasks.apps.TasksConfig',
 'tickets.apps.TicketsConfig']
Installed Middleware:
['tenants.middleware.main.TenantMainMiddleware',
 'debug_toolbar.middleware.DebugToolbarMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
[...]

PR #337, opened 2019-10-08
Save should not raise `ValueError` on invalid phonenumber

### LINKED ISSUE #334: get_prep_value() on the field SHOULD NOT trigger ValueError
Hi,

Django==1.11.23 here.

I have a signal set to update a model when a related model changes. When the model containing the phonenumber contains a previously valid, now not valid phonenumber, the save will trigger a `ValueError`, thus crashing django and hindering the save of the original object.

Saving the model ***SHOULD NOT*** raise an exception at all. If the data can be saved, just save it, don't complain. Data validation as per the Django framework standard happens in the `clean()` method, not in the save method.

You are crashing probably a lot of applications now, with the changed `get_prep_value()`.

Please remove the `ValueError`s from `get_prep_value()`.

Here's a traceback that will show how you crash a save from a related model save:
```python
File "/home/project/backend/project/account/signals.py" in handle_user_save
  15.         related_model.save()

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/base.py" in save
  808.                        force_update=force_update, update_fields=update_fields)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/base.py" in save_base
  838.             updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/base.py" in _save_table
  905.                                       forced_update)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/base.py" in _do_update
  955.         return filtered._update(values) > 0

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/query.py" in _update
  667.         return query.get_compiler(self.db).execute_sql(CURSOR)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/sql/compiler.py" in execute_sql
  1204.         cursor = super(SQLUpdateCompiler, self).execute_sql(result_type)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/sql/compiler.py" in execute_sql
  876.             sql, params = self.as_sql()

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/sql/compiler.py" in as_sql
  1170.                 val = field.get_db_prep_save(val, connection=self.connection)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/fields/__init__.py" in get_db_prep_save
  770.                                       prepared=False)

File "/home/project/venv/lib/python3.7/site-packages/django/db/models/fields/__init__.py" in get_db_prep_value
  762.             value = self.get_prep_value(value)

File "/home/project/venv/lib/python3.7/site-packages/phonenumber_field/modelfields.py" in get_prep_value
  83.                 raise ValueError("“%s” is not a valid phone number." % value.raw_input)
```

This is so wrong on many levels. Please fix this ASAP.

karolyi:
Save() should only raise `ValueError`s when the data to be stored
is incompatible with the raw field data. A string that contains
data that is not a valid phone number doesn't count as invalid,
and should be checked in clean(), not in save().

closes #334

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ downloads
 eggs
 parts
 build
+venv
 *.egg-info
 *.egg
 .DS_Store
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -213,14 +213,6 @@ class PhoneNumberFieldTestCase(TestCase):
         m = models.MandatoryPhoneNumber.objects.defer("phone_number").get(pk=m.pk)
         self.assertEqual(m.phone_number, self.test_number_1)
 
-    def test_filter_invalid(self):
-        models.NullablePhoneNumber.objects.create(phone_number=self.test_number_1)
-        models.NullablePhoneNumber.objects.create(phone_number="")
-        models.NullablePhoneNumber.objects.create(phone_number=None)
-        with self.assertRaises(ValueError) as cm:
-            models.NullablePhoneNumber.objects.filter(phone_number="123").exists()
-        self.assertEqual(force_text(cm.exception), "“123” is not a valid phone number.")
-
     def test_filter_int(self):
         models.NullablePhoneNumber.objects.create(phone_number=self.test_number_1)
         models.NullablePhoneNumber.objects.create(phone_number="")
@@ -245,38 +237,6 @@ class PhonenumerFieldAppTest(TestCase):
             models.TestModel.objects.all(),
             [(tm.pk, "", "+41524242424")],
             transform=phone_transform,
-        )
-
-    def test_create_invalid_number(self):
-        with self.assertRaises(ValueError) as cm:
-            models.TestModel.objects.create(phone="invalid")
-        self.assertEqual(
-            force_text(cm.exception), "“invalid” is not a valid phone number."
-        )
-
-    def test_save_invalid_number(self):
-        tm = models.TestModel.objects.create(phone="+1 604-333-4444")
-        tm.phone = "invalid"
-        with self.assertRaises(ValueError) as cm:
-            tm.save()
-        self.assertEqual(
-            force_text(cm.exception), "“invalid” is not a valid phone number."
-        )
-
-    def test_save_update_field_invalid_number(self):
-        tm = models.TestModel.objects.create(phone="+1 604-333-4444")
-        tm.phone = "invalid"
-        with self.assertRaises(ValueError) as cm:
-            tm.save(update_fields=["phone"])
-        self.assertEqual(
-            force_text(cm.exception), "“invalid” is not a valid phone number."
-        )
-
-    def test_update_to_invalid_number(self):
-        with self.assertRaises(ValueError) as cm:
-            models.TestModel.objects.update(phone="invalid")
-        self.assertEqual(
-            force_text(cm.exception), "“invalid” is not a valid phone number."
         )
 
     def test_save_blank_phone_to_database(self):
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -1,6 +1,5 @@
 from django.conf import settings
 from django.core import checks
-from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.encoding import force_text
 from django.utils.translation import gettext_lazy as _
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.core import checks
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.encoding import force_text
 from django.utils.translation import gettext_lazy as _
@@ -80,9 +81,6 @@ class PhoneNumberField(models.CharField):
         if value:
             if not isinstance(value, PhoneNumber):
                 value = to_python(value)
-
-            if not value.is_valid():
-                raise ValueError("“%s” is not a valid phone number." % value.raw_input)
 
             format_string = getattr(settings, "PHONENUMBER_DB_FORMAT", "E164")
             fmt = PhoneNumber.format_map[format_string]


karolyi:
LGTM.

tests/tests.py
```diff
@@ -247,38 +239,6 @@ def test_save_field_to_database(self):
             transform=phone_transform,
         )
 
-    def test_create_invalid_number(self):
```
jdufresne: Rather than removing these tests, please alter them to assert the new behavior and side effects.

.gitignore
```diff
@@ -9,6 +9,7 @@ downloads
 eggs
 parts
 build
+venv
```
jdufresne: Please avoid unrelated changes to the issue being addressed.

tests/tests.py
karolyi: Did you have a look at the removed tests? They were asserting the exceptions, which I was explicitly emphasizing that it SHOULD NOT raise.

.gitignore
karolyi: venv is a common name for a virtualenv.

jdufresne: I'm not saying otherwise.

I'm saying don't mix non-functional code changes in a functional PR. Please keep the change targeted to the issue at hand. If you want suggest other changes, please do so as separate commits so we aren't reviewing multiple changes together.

tests/tests.py
jdufresne: Understood. 

If we're going to change the existing behavior, we should now assert the expected behavior when saving an invalid number. 

When saving an invalid number, the side effect is ... ?

The test should answer this question.
So if we behavior change the behavior from raise an exception

jdufresne:
> A string that contains data that is not a valid phone number doesn't count as invalid

I'm not sure I agree with this statement. An invalid phone number should not be saved by this field. `ValueError` occur for other field types when invalid data is saved. For example, trying to save `'abc'` to an `IntegerField` results in:

```
ValueError: invalid literal for int() with base 10: 'abc'
```

jdufresne:
Just a s followup, this has come up in the past, here is the discussion from last time:

https://github.com/stefanfoulis/django-phonenumber-field/issues/306#issuecomment-501866846

karolyi:
> > A string that contains data that is not a valid phone number doesn't count as invalid
> 
> I'm not sure I agree with this statement. An invalid phone number should not be saved by this field. `ValueError` occur for other field types when invalid data is saved. For example, trying to save `'abc'` to an `IntegerField` results in:
> 
> ```
> ValueError: invalid literal for int() with base 10: 'abc'
> ```

Dude, look at #334. My entire reasoning against raising exceptions on invalid phonenumbers is there. In a nutshell (again), a check on phonenumber **SHOULD** happen in the clean() method **BEFORE** saving, and save should only raise when the **RAW** data doesn't fit into the DB field's data type. This is the Django standard, not something invented by me. See my use case for an error it raises, and in the issue I detail as to why you're probably breaking a lot of appliations now.

I had to create a painful workaround because of this issue.

jdufresne:
Thanks for reporting the issue and opening the PR. However, I'm going to stay with the decision made in #306. As other Django fields raise `ValueError` (such as `IntegerField` when passed a non-`int`), I think this is in line with Django semantics.  If you have a link to definitive docs that say this is incorrect, I'll reconsider and take another look.

karolyi:
Did you have a look at my issue? You are breaking most probably a lot of applications now.

Like I said (this is the 3rd time): if the value fits, that is it's a string, it SHOULD NOT raise any error, just save silently.

If this is not enough for you, I don't know what is: https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.clean

karolyi:
@stefanfoulis can you please add your 2 cents here as well?

stefanfoulis:
@karolyi I understand your reasoning and the frustration it must cause when code breaks because of this seemingly unrelated change in django-phonenumber-field.

The problem we are trying to solve is avoiding invalid data reach the database. Also to avoid the problem of needing to have an instance of the `PhoneNumber` wrapper with an invalid value. However, as you've shown, we can't really avoid that for all cases anyway. A setting may be changed or a new rule added to the upstream library which changes the definition of what a valid number is after it has already been saved to the database.
We've seen problems in the past where it was not noticed that the number is invalid because it was assigned and the raw (invalid) string saved without errors.

I think we need to take a step back and define the correct behaviour in a few different situations so we really end up with consistent and correct behaviour. 

🤔 I need to think about this a bit more. I am leaning towards saving a model where we never did anything to the phone field should work, even if it is invalid.

Here are some situations we should be clear on what the correct behaviour is:
(I'll think about more later. Please add your own if you think about other significant situations.)

Allow saving with an invalid number that is already stored in the database (valid string)?

```
>>> contact = Contact.objects.get(id=123)
>>> contact.save()  # no errors
>>> contact.clean()  # raise ValidationError

# but what about this?
>>> repr(contact.phone)
# perhaps something like InvalidPhoneNumber("+41 invalid")?
```

What should this do?

```
contact = Contact()
contact.phone_number = "invalid phone number"  # raise ValidationError or silently create a `InvalidPhoneNumber` wrapper?
contact.phone_number = 123  # raise ValueError?
```

more...

karolyi:
Thanks for the concise and well thought out response. Let me give some more insight too:

As I pointed out before, the [Django standard is to validate data in the `clean()` method](https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.clean), and raise `ValidationError`s there. The container field should only raise `ValueError`s when the python data structure doesn't fit the underlying field; e.g. putting an integer value into a `Charfield`. Since `PhoneNumberField` uses a `CharField` underneath, that functionality is already a given.

Assigning a value into the field shouldn't raise an error as well, as that would lead to unexpected exceptions in code that already uses your module, which is what we're trying to avoid here.

However, I don't have a problem with displaying informational error messages on the `repr` usage. In fact I like the idea of that, since it points out to developers that something's not right with the record they're trying to use.

All I want is to revert back to painless functionality. I've built a workaround around my case that sends emails whenever a previously valid, now invalid phone number gets saved and an error would be raised which would crash my code. I get surprisingly lots of emails because of this, hence my assumption that the current functionality breaks probably a lot of code using this module.

The consequences of this error can be vast: in my case, the resaving of the model from a signal should also update a xapian search db, what doesn't happen because of the exception raised, and thus corrupts the search database for my project. That is, it happened until I built my workaround, that is really painful: do a `Model.objects.filter(pk=model.pk).update(phonenumber=phonenumber)`, and then update the search DB. This avoids the use of `save()` and updates the model even if the phone number is invalid. I think now you get the idea as to why this is a dangerous change, and why I'm so opposed to it.

karolyi:
can we pick up some action here, please?

karolyi:
bump

karolyi:
bump

stefanfoulis:
@karolyi I created a new pull request that includes your removal of `ValueError` and some more changes regarding the handling of invalid numbers and some more tests: #346

stefanfoulis:
replaced by #346

PR #346, opened 2019-11-23
Don't raise ValueError in get pre value (and more improvements)

stefanfoulis:


phonenumber_field/modelfields.py
```diff
@@ -77,16 +77,24 @@ def get_prep_value(self, value):
...
+        if not value:
+            return super().get_prep_value(value)
 
-            if not value.is_valid():
-                raise ValueError("“%s” is not a valid phone number." % value.raw_input)
+        if isinstance(value, PhoneNumber):
+            parsed_value = value
+        else:
+            # Convert the string to a PhoneNumber object.
+            parsed_value = to_python(value)
```
karolyi: Not sure if this line won't raise an error in case of an invalid number, but other than this, LGTM.

stefanfoulis: As long as phonenumbers does as promised and raises a `NumberParseException` if something is wrong with the number we should be fine. In `to_python` all numbers that cause a `NumberParseException` are stored in their raw format.
https://github.com/stefanfoulis/django-phonenumber-field/blob/cdb3d1251a0c4de0f889a249c141e20aec167c7e/phonenumber_field/phonenumber.py#L115-L117

stefanfoulis:
I'll release this soon-ish with and bump the version to `4.0` because it is a significant change in behaviour of invalid numbers.

stefanfoulis:
Version `4.0.0` released

karolyi:
Great stuff, thank you. I'll install it and so it will get thorough testing (the users are the best testers, as the saying goes).

Just a hint, I wouldn't upgrade major versions that aggressively. If you intend to follow a semver-like versioning, major version changes signal that there are breaking API changes, but in this case there aren't.

PR #349, opened 2020-01-09
Fix initial value problem for PhonePrefixSelect and default use PHONE…

### LINKED ISSUE #202: initialize PhoneNumberPrefixWidget with initial don't make effect

rayer4u:
Another solution to fix #202

### DIFF
--- a/phonenumber_field/widgets.py
+++ b/phonenumber_field/widgets.py
@@ -21,6 +21,9 @@ class PhonePrefixSelect(Select):
         language = translation.get_language() or settings.LANGUAGE_CODE
         if language:
             locale = Locale(translation.to_locale(language))
+            if not initial:
+                region = getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
+                initial = region
             for prefix, values in _COUNTRY_CODE_TO_REGION_CODE.items():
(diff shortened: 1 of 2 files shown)


rayer4u:
update with new code.

Guilouf:
Why is this PR not merged yet ? Initial values is quite an important feature

jdufresne:
Thanks! merged as 29b4866c06cfce3edafd813365d8f32c28efe499.

PR #353, opened 2020-02-16
Use extras in installation docs

### LINKED ISSUE #350: Install phonenumbers dependency automatically?
The instruction currently say to explicitly `pip install phonenumbers`, but could this not just be declared as a dependency in setup.py so pip will automatically install it when installing `django-phonenumber-field`?

jdufresne:
Fixes #350

@matthijskooijman would you like to take a look?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,13 +32,15 @@ Installation
 
 ::
 
-    pip install django-phonenumber-field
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Looks good, thanks!

PR #370, opened 2020-06-10
Update setup.py

simantabaral:
fixed error


jdufresne:
What is the purpose of this change? What is it solving? AFAICT, setuptools doesn't have a setup2 function.

PR #391, opened 2020-10-25
Added dependency to setup.py

iamoracle:
Missing dependency `phonenumbers`. I had to install it manually.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,6 @@
 from setuptools import setup
 
-setup()
+setup(
+  install_requires = [
+    'phonenumbers',
+  ])


francoisfreitag:
Hi #iamoracle,

The [installation section](https://github.com/stefanfoulis/django-phonenumber-field#installation) between provides a choice between `phonenumbers` and `phonenumberslite`, that’s why the dependency is not listed as an `install_requires`. With the suggested change, `phonenumbers` would always be installed.

To hint users for the dependency, both are listed as `extra_requires`.

francoisfreitag (changes requested):
See above comment.

francoisfreitag:
Closing for inactivity.

Issue #382, opened 2020-08-17
Does django-phonenumber-field 5.0 support Django 2.2?


gopackgo90:
According to the [release notes](https://github.com/stefanfoulis/django-phonenumber-field/blame/5.0.0/CHANGELOG.rst#L18):

> * Drop support for end-of-life Django 1.11 and 2.2.

But it looks like in the [Travis configuration](https://github.com/stefanfoulis/django-phonenumber-field/blob/5.0.0/.travis.yml#L10-L17), Django 2.2 is still supported and, in fact, [it was Django 2.1, not 2.2, that was removed](https://github.com/stefanfoulis/django-phonenumber-field/commit/7fed681f8c644825af8191705c033179536f3f99#diff-354f30a63fb0907d4ad57269548329e3).

Could you clarify this? It looks like it's just a typo in the 5.0 release notes and Django 2.1 is now unsupported, but Django 2.2 is still supported, is that correct?

adamrparsons:
Following too because I've noticed that Pipenv updated us to 5.0 while we use Django 2.2 LTS. Haven't noticed any issues yet but I haven't had time to test it thoroughly.

StevenMapes:
It's a mistake in the release notes that **really** needs correcting. If you look in the setup.cfg file you'll see that it's 2.2+ that is needed ```install_requires = Django >= 2.2``` so the readme should say that it's dropped support < Django 2.2 as that certainly isn't EOL

PR #392, opened 2020-10-25
Update CHANGELOG.rst - Fix for issue #382

StevenMapes:
Corrected the change log for 5.0.0 which incorrectly said it dropped support for Django 2.2 stating it was EOL when it's not EOL as it's an LTS release. This is confirmed in the setup files.


stefanfoulis (approved):
🙈

PR #399, opened 2021-01-16
Drop support for Python 3.5

francoisfreitag:
Python 3.5 end of life was Sept 5th, 2020.
https://www.python.org/downloads/release/python-3510/


stefanfoulis (approved):
🚀

PR #400, opened 2021-01-16
Replace deprecated force_text with force_str

francoisfreitag:
The `force_text` helper will be removed in Django 4.0.

https://github.com/django/django/commit/d55e88292723764a16f0689c73bc7e739dfa6047


stefanfoulis (approved):
🚀

PR #406, opened 2021-01-17
Use default Python version to finish coveralls build

francoisfreitag:
The python version does not matter for that build, use the default.


stefanfoulis (approved):
🚀

PR #407, opened 2021-01-21
Don’t package the .github directory

francoisfreitag:
It contains CI configuration for the project and should not be
distributed.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,3 +3,5 @@ include README.rst
 include tox.ini
 recursive-include phonenumber_field/locale django.mo django.po
 recursive-include tests *.py
+
+prune .github


stefanfoulis:
weird 🤔
I am pretty sure that it only included files in the package that were explicitly included in `MANIFEST.in`. Something must have changed along the line.

stefanfoulis:
Maybe because of `include_package_date=True`: https://github.com/stefanfoulis/django-phonenumber-field/blob/main/setup.cfg#L33 (https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html).
As far as I understand it includes everything that is under source-control 🤔 .

Ideally we'd explicitly list which directories/files we want to include.

francoisfreitag:
No, you are correct. The `.github` and `.gitignore` (#408) were added by `setuptools_scm` (#409).

PR #408, opened 2021-01-21
Don’t package .gitignore

francoisfreitag:
File is specific to version control and should not be distributed.


francoisfreitag:
No, you are correct. The `.github` (#407) and `.gitignore` were added by `setuptools_scm` (#409).

Issue #224, opened 2018-07-12
join jazzband


stefanfoulis:
Lets join [jazzband](https://jazzband.co/) for better maintainability and to reach more potential contributors.

The [Guidelines](https://jazzband.co/about/guidelines) describe the requirements to join.

Improvements needed:
- [ ] docs (https://jazzband.co/about/guidelines#documentation)
- [ ] packaging/versioning: setup so that semi-automatic releasing can work on pypi trough travis (https://jazzband.co/about/releases#packaging)

Final Prepping:
- [ ] add ``jazzband`` user on pypi for releases (https://jazzband.co/about/releases)
- [ ] CONTRIBUTING.md (https://jazzband.co/about/guidelines#contributing-guidelines)
- [ ] Badge (https://jazzband.co/about/guidelines#badges)

joshua-s:
Any progress on this?

francoisfreitag:
Assigning to Stefan, the code should be ready now, but https://readthedocs.org/ needs to be configured.

francoisfreitag:
https://github.com/jazzband/help/issues/310

PR #409, opened 2021-01-21
Use setuptools_scm to get the version number

francoisfreitag:
Follow recommendation from https://jazzband.co/about/releases#packaging.

Refs #224


francoisfreitag:
The `.github` (#407) and `.gitignore` were added by `setuptools_scm` (#409).

PR #416, opened 2021-02-13
Simplify tox environments

francoisfreitag:
The Python version is now dictated by the github action configuration
file.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,10 +3,10 @@ envlist =
     black
     flake8
     isort
-    py{36,37,38}-dj22
-    py{36,37,38}-dj30
-    py{36,37,38}-dj31
-    py{36,37,38}-djmaster
+    dj22
+    dj30
+    dj31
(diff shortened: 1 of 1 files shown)


stefanfoulis (approved):
ok, so if I understand correctly tox won't do any matrix builds regarding python version anymore then? That'll be done by github actions?
 
(I'll need to remember to manually pick python versions when running tests locally)

francoisfreitag:
The CI uses the GitHub actions configuration, the build matrix is the cartesian product of environments declared in https://github.com/stefanfoulis/django-phonenumber-field/blob/4816c5000bc84f2c37698f125e26e85bb9dbe221/.github/workflows/test.yml#L15-L23

We can leave the build matrix repeated in `tox.ini`. I rely on CI for testing across the matrix, and locally use my system version of Python. That’s why it didn’t seem useful to me and I suggested removing it.

PR #426, opened 2021-02-19
Prefer type(self) over self.__class__

francoisfreitag:
Use the public interface to retrieve the class.


francoisfreitag:
I personally prefer it, it’s public interface. Using __class__ is fine as well.

PR #437, opened 2021-04-16
Lazy load form field error message translation

francoisfreitag:
Follow pattern from Django form fields. Since fields are loaded at
import time, prevents calling into the translation machinery before
Django has completed its setup.

Thanks traviselam for reporting.


th3hamm0r:
@francoisfreitag What is the status of this PR? It would be nice to get this into a release!

th3hamm0r:
Ok, I've just tested your fix and the it seems like due to the call of `error_message.format(...)` the lazy translation gets evaluated again on import time. Maybe that's the reason the PR hasn't been merged(?).

francoisfreitag:
Hi, thanks for raising this. The PR needs a review before to be integrated. Pinging @stefanfoulis :slightly_smiling_face:

francoisfreitag:
> So unfortunately this fix is still not working.

Tight. Thanks for reviewing!
I’ll update the fix to actually defer the translation.

francoisfreitag:
@th3hamm0r Could you confirm the new fix actually fixes the issue for you this time?

th3hamm0r:
@francoisfreitag Awesome, that fix works! Thanks for the fast response!

stefanfoulis (approved):
🚀

PR #439, opened 2021-04-21
Remove intermediary assignment in PhonePrefixSelect

francoisfreitag:
`region` is not reused, can directly assign to the `initial` variable.


stefanfoulis (approved):
🚀

PR #445, opened 2021-05-20
Ignore coverage reports

francoisfreitag:
Running tox locally generates a coverage-reports directory. It is the
result of a build and can safely be ignored.


thedrow (approved):
Thanks!

PR #447, opened 2021-05-20
Use older Django compiled messages

francoisfreitag:
The compiled messages on the main branch have a different format from
older Django. Recompile them with Django 3.1, that format is stable
across versions.


stefanfoulis (approved):
🚀

PR #454, opened 2021-07-28
fix: update Czech translation files

melanger:
add a missing translation string and regenerate files


francoisfreitag (approved):
Sweet, thanks!

PR #456, opened 2021-08-04
Update django.po

willunicamp:
Translated main error message in pt_BR


francoisfreitag:
Thanks for translating! :tada:

PR #467, opened 2021-09-27
Extending PrefixSelectWidget with Allowed, Excluded, & Preferred settings

rolavides:
Extending PrefixSelectWidget with Allowed, Excluded, & Preferred settings.

This provides a way to:

- list only the countries the package user supports
- exclude countries where package user's service still do not support
- "float" a list of country above the alphabetically sorted list.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -111,6 +111,45 @@ Choices:
 - ``"NATIONAL"``,
 - ``"RFC3966"``.
 
+``PHONENUMBER_EXCLUDED_REGIONS``
+------------------------------
+
+A list of `ISO-3166-1 <https://en.wikipedia.org/wiki/ISO_3166-1#Current_codes>`_
+two-letter country code to exclude from the Prefix Select Widget.
+
+Supersedes all the other Prefix Select Widget display settings. 
+
+Default: ``None``.
+
+``PHONENUMBER_ALLOWED_REGIONS``
+------------------------------
+
+A list of `ISO-3166-1 <https://en.wikipedia.org/wiki/ISO_3166-1#Current_codes>`_
+two-letter country code to exclude from the Prefix Select Widget.
+
+Useful in cases where not all countries are supported yet. Limits SMS sending or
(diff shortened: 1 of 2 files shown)


francoisfreitag:
Hi,
Thanks for your contribution. It may be useful to some users :+1:.

Django tries hard to avoid adding settings, I think this project should follow. Why not drop the settings? Users who want to have a default `PhonePrefixSelect` can subclass and specify the list of countries in their subclass, without settings.

francoisfreitag:
Could you add an entry to the [changelog](https://github.com/stefanfoulis/django-phonenumber-field/blob/main/CHANGELOG.rst) as well?

rolavides:
Hi, 

Thanks for looking into the PR.

> Django tries hard to avoid adding settings, I think this project should follow. Why not drop the settings? 

I think the usefulness of these addition is in the context of having all PhoneNumberField be presented and behave the same in the entirety of the application, without losing the flexibility of having a form deviating from the application configuration.

> Users who want to have a default `PhonePrefixSelect` can subclass and specify the list of countries in their subclass, without settings.

I agree, we can always just have our own custom widget or reimplement the widget.

---

I'm giving the use-cases of these addition a second look. I'll close the PR for now. There are also some things that needed additional checks added before I should've made a PR.

Should I attempt another PR with a changelog, how should the version number be bumped for this specific set of changes @francoisfreitag?

francoisfreitag:
> I think the usefulness of these addition is in the context of having all PhoneNumberField be presented and behave the same in the entirety of the application, without losing the flexibility of having a form deviating from the application configuration.

That can be achieved by subclassing the field for the needs of your project. It’s about the same amount of configuring but better localized. Having the application change behavior based on a global element is usually not a good idea.

---

> Should I attempt another PR with a changelog, how should the version number be bumped for this specific set of changes?

A new PR or reopening this PR is fine.
Don’t worry about the version number, I’ll bump it when I make a release. Just add your changes above the 5.2.0 section.

PR #480, opened 2021-12-20
Remove SonarCloud build

francoisfreitag:
The build has been red for a good year, with nobody volunteering to fix
it. Looking at the reported issues, they are considered code smells but
not particularly interesting:

- A test uses Exception, it recommends using a more specific subclass.
  (considered major)
- It recommends extracting a test string (phone number) that is reused
  in multiple tests into a variable. (considered critical)

IMO, both these issues are in the very minor bucket and not worth
fixing.


stefanfoulis (approved):
🚀

PR #482, opened 2022-01-06
Drop support for Python 3.6

francoisfreitag:
This Python version has reached end of life and won’t receive any fixes.
https://www.python.org/downloads/


thedrow (approved):

thedrow:
Would you mind running [pyupgrade](https://github.com/asottile/pyupgrade) as well with `--py37-plus`?

PR #484, opened 2022-01-11
Prefer f-string to format strings

francoisfreitag:


phonenumber_field/phonenumber.py
```diff
@@ -43,9 +43,7 @@ def __str__(self):
 
     def __repr__(self):
         if not self.is_valid():
-            return str(
-                "Invalid{}(raw_input={})".format(type(self).__name__, self.raw_input)
-            )
+            return str(f"Invalid{type(self).__name__}(raw_input={self.raw_input})")
```
thedrow, line 46: Do we really need to convert to a string?

thedrow (changes requested):
Looks good but I have one question.

phonenumber_field/phonenumber.py
francoisfreitag: Addressed in #485.

thedrow (approved):

PR #490, opened 2022-02-16
Remove `maxlength` attribute for html5 compliance

sterliakov:
Pages that use `PhonePrefixSelect` do not validate with [vnu validator](https://validator.github.io/validator/) because `<select>` tag is rendered with `maxlength` attribute which makes no sense for that tag. With this PR we just inspect context and pop `maxlength` attribute if it is present.

### DIFF
--- a/phonenumber_field/widgets.py
+++ b/phonenumber_field/widgets.py
@@ -43,6 +43,8 @@ class PhonePrefixSelect(Select):
         super().__init__(choices=sorted(choices, key=lambda item: item[1]))
 
     def get_context(self, name, value, attrs):
+        attrs = (attrs or {}).copy()
+        attrs.pop("maxlength", None)
         return super().get_context(name, value or self.initial, attrs)
 
 
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -33,6 +33,12 @@ class PhoneNumberPrefixWidgetTest(SimpleTestCase):
             '<select name="_0"><option value="" selected>---------</option>', rendered
         )
 
+    def test_maxlength_not_in_select(self):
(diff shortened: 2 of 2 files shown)


francoisfreitag:
Please add a test that prevents the issue from commit back.

amateja:
Hi @sterliakov,

Your issue is caused by `PhoneNumberPrefixWidget` parent class - `django.forms.widgets.MultiWidget`. When `MultiWidget.get_context` is called it also calls subwidgets' `get_context` methods sharing own context containing `maxlength` attribute with them.

Instead modifying context of the widget I would recommend to strip `maxlength` key from `attrs` like this:
```python
    def get_context(self, name, value, attrs):
        attrs = (attrs or {}).copy()
        attrs.pop('maxlength', None)
        return super().get_context(name, value or self.initial, attrs)
```
And in terms of testing we should simply check if `maxlength` keyword is not part of `<select>` html tag. We could do this in `PhoneNumberPrefixWidgetTest` class like this:

```python
    def test_maxlength_not_in_select(self):
        widget = PhoneNumberPrefixWidget()
        html = widget.render(name="widget", value=None, attrs={"maxlength": 32})
        self.assertIn('<select name="widget_0">', html)
```
I've done my own tests and [W3C validator](https://validator.w3.org/) was happy with the changes.

Regards,
Andrzej

sterliakov:
Hello @amateja, I apologize for such a late response. I applied changes requested changes.

Please tell if I can be useful for this or other issues - I'm willing to help:)

amateja (approved):

amateja:
Thanks for your contribution.

Issue #461, opened 2021-08-11
form field clears when either country code or phone number text is empty


KunjPrasad:
For an edge case where user would have entered phone number text but not select country code, or selected the country code but not added the phone number, I am noticing that when the form is submitted, then due to [`if all(values):`](https://github.com/stefanfoulis/django-phonenumber-field/blob/main/phonenumber_field/widgets.py#L74) check in `PhoneNumberPrefixWidget.value_from_datadict` method, the rerendered form has user entries wiped out.

I feel a better experience would be to leave the partial user data on form as-is. I am not sure if the criteria is even helping at all, and if it is better removed, i.e. simply `return "%s.%s" % tuple(values)` for all cases. Anyways, the `to_python` method in `formfields.PhoneNumberField` will raise a validation error ([here](https://github.com/stefanfoulis/django-phonenumber-field/blob/main/phonenumber_field/formfields.py#L46))

amateja:
This should be addressed as a side effect of implementing #493 and is related to #489.

Issue #458, opened 2021-08-09
How do you limit the number of choices of the widget


superflyer11:
Is there a built-in setting to limit the number of choices to specific ones? Or do I need to override the select widget?

amateja:
Currently there is no other way than a hacky one. After instantiation of `PhoneNumberPrefixWidget` you might access `PhoneNumberPrefixWidget.widgets` property, get an instance of `PhonePrefixSelect` and alter `PhonePrefixSelect.choices` property.

On the other hand we might change both `PhoneNumberPrefixWidget` and `PhonePrefixSelect` constructor to accept `choices` argument... but I need to get some thumbs up on it.

PR #493, opened 2022-03-02
PhoneNumberPrefixWidget improvement for regions sharing same country prefix

### LINKED ISSUE #133: PhoneNumberPrefixWidget sometimes selects wrong country as initial value

### LINKED ISSUE #251: Optionally store region in database
Number prefixes are not a 1:1 mapping with regions. When using the widget with region selection this may cause confusing behaviour (see #133 ).

Lets save the region in the db (either in the same `CharField` with special formatting or in a separate field).

See https://github.com/stefanfoulis/django-phonenumber-field/issues/133#issuecomment-230028890 as a starting-point.

### LINKED ISSUE #489: Using default country on form errors
Hello, first of all thanks for this application, I found an issue when adding a wrong number, I selected `United States +1` and after the form is submitted and the page is re-loaded, I see the selection is `American Samoa +1`, inspecting the code shows the code `+1` is not unique which makes the HTML show the first element with `+1` value, is there any way to fix that issue?

image

This is the setting I am using:

```python
PHONENUMBER_DEFAULT_REGION = "US"
```

amateja:
As promised in #251 this PR improves `PhoneNumberPrefixWidget` handling for regions sharing the same country prefix. Current behavior does not follow requirements for `Select.Choices`. Choices should be indexed by [unique](https://docs.djangoproject.com/en/4.0/ref/models/fields/#enumeration-types) values and this requirement was not met for 12 country prefixes.

Fixes #133
Fixes #251
Fixes #489

### DIFF
--- a/phonenumber_field/widgets.py
+++ b/phonenumber_field/widgets.py
@@ -10,36 +10,47 @@ from phonenumbers.phonenumberutil import (
     region_code_for_number,
 )
 
-from phonenumber_field.phonenumber import PhoneNumber
+from phonenumber_field.phonenumber import PhoneNumber, to_python
 
 try:
     import babel
 except ModuleNotFoundError:
     babel = None
 
+# ISO 3166-1 alpha-2 to national prefix
+REGION_CODE_TO_COUNTRY_CODE = {
+    region_code: country_code
+    for country_code, region_codes in COUNTRY_CODE_TO_REGION_CODE.items()
+    for region_code in region_codes
+}
+
+
+def localized_choices(language):
+    if babel is None:
+        raise ImproperlyConfigured(
+            "The PhonePrefixSelect widget requires the babel package be installed."
+        )
+
+    choices = [("", "---------")]
+    locale_name = translation.to_locale(language)
+    locale = babel.Locale(locale_name)
+    for region_code, country_code in REGION_CODE_TO_COUNTRY_CODE.items():
+        region_name = locale.territories.get(region_code)
+        if region_name:
+            choices.append((region_code, f"{region_name} +{country_code}"))
+    return choices
+
 
 class PhonePrefixSelect(Select):
     initial = None
 
     def __init__(self, initial=None):
-        if babel is None:
-            raise ImproperlyConfigured(
-                "The PhonePrefixSelect widget requires the babel package be installed."
-            )
-
-        choices = [("", "---------")]
         language = translation.get_language() or settings.LANGUAGE_CODE
-        locale = babel.Locale(translation.to_locale(language))
-        if not initial:
+        choices = localized_choices(language)
+        if initial is None:
             initial = getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
-        for prefix, values in COUNTRY_CODE_TO_REGION_CODE.items():
-            prefix = "+%d" % prefix
-            if initial and initial in values:
-                self.initial = prefix
-            for country_code in values:
-                country_name = locale.territories.get(country_code)
-                if country_name:
-                    choices.append((prefix, f"{country_name} {prefix}"))
+        if initial in REGION_CODE_TO_COUNTRY_CODE:
+            self.initial = initial
         super().__init__(choices=sorted(choices, key=lambda item: item[1]))
 
     def get_context(self, name, value, attrs):
@@ -60,22 +71,28 @@ class PhoneNumberPrefixWidget(MultiWidget):
         super().__init__(widgets, attrs)
 
     def decompress(self, value):
-        if value:
-            if isinstance(value, PhoneNumber):
-                if value.country_code and value.national_number:
-                    return [
-                        "+%d" % value.country_code,
-                        national_significant_number(value),
-                    ]
-            else:
-                return value.split(".")
-        return [None, ""]
+        if isinstance(value, PhoneNumber):
+            if not value.is_valid():
+                region = getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
+                region_code = getattr(value, "_region", region)
+                return [region_code, value.raw_input]
+            region_code = region_code_for_number(value)
+            national_number = national_significant_number(value)
+            return [region_code, national_number]
+        return [None, None]
 
     def value_from_datadict(self, data, files, name):
-        values = super().value_from_datadict(data, files, name)
-        if all(values):
-            return "%s.%s" % tuple(values)
-        return ""
+        region_code, national_number = super().value_from_datadict(data, files, name)
+
+        if national_number is None:
+            national_number = ""
+        number = to_python(national_number, region=region_code)
+        if isinstance(number, str):
+            number = PhoneNumber()
+            number.raw_input = national_number
+        if not number.is_valid():
+            setattr(number, "_region", region_code)
+        return number
 
 
 class PhoneNumberInternationalFallbackWidget(TextInput):
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -1,8 +1,10 @@
+import phonenumbers
+from django import forms
 from django.core.exceptions import ImproperlyConfigured
 from django.test import SimpleTestCase, override_settings
 from django.utils import translation
 
-from phonenumber_field import widgets
+from phonenumber_field import formfields, widgets
 from phonenumber_field.phonenumber import PhoneNumber
 from phonenumber_field.widgets import (
     PhoneNumberInternationalFallbackWidget,
@@ -25,22 +27,47 @@ class PhonePrefixSelectTest(SimpleTestCase):
         super().tearDown()
 
 
+def example_number(region_code: str) -> PhoneNumber:
+    number = phonenumbers.example_number(region_code)
+    e164 = phonenumbers.format_number(number, phonenumbers.PhoneNumberFormat.E164)
+    return PhoneNumber.from_string(e164, region=region_code)
+
+
+EXAMPLE_NUMBERS = (
+    example_number("US"),
+    example_number("CA"),
+    example_number("GB"),
+    example_number("GG"),
+    example_number("PL"),
+    example_number("IT"),
+)
+
+
 class PhoneNumberPrefixWidgetTest(SimpleTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        class TestForm(forms.Form):
+            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)
+
+        cls.form_class = TestForm
+
     def test_initial(self):
         rendered = PhoneNumberPrefixWidget(initial="CN").render("", "")
         self.assertIn('<option value="">---------</option>', rendered)
-        self.assertIn('<option value="+86" selected>China +86</option>', rendered)
+        self.assertIn('<option value="CN" selected>China +86</option>', rendered)
 
     @override_settings(PHONENUMBER_DEFAULT_REGION="CN")
     def test_uses_default_region_as_initial(self):
         rendered = PhoneNumberPrefixWidget().render("", "")
         self.assertIn('<option value="">---------</option>', rendered)
-        self.assertIn('<option value="+86" selected>China +86</option>', rendered)
+        self.assertIn('<option value="CN" selected>China +86</option>', rendered)
 
     def test_no_initial(self):
         rendered = PhoneNumberPrefixWidget().render("", "")
         self.assertIn('<option value="" selected>---------</option>', rendered)
-        self.assertIn('<option value="+86">China +86</option>', rendered)
+        self.assertIn('<option value="CN">China +86</option>', rendered)
 
     @override_settings(USE_I18N=True)
     def test_after_translation_deactivate_all(self):
@@ -48,6 +75,128 @@ class PhoneNumberPrefixWidgetTest(SimpleTestCase):
         rendered = PhoneNumberPrefixWidget().render("", "")
         self.assertIn(
             '<select name="_0"><option value="" selected>---------</option>', rendered
+        )
+
+    def test_value_from_datadict(self):
+        tests = [
+            (
+                {
+                    "phone_0": phonenumbers.region_code_for_number(number),
+                    "phone_1": phonenumbers.national_significant_number(number),
+                },
+                number,
+            )
+            for number in EXAMPLE_NUMBERS
+        ]
+
+        for data, expected in tests:
+            with self.subTest(data):
+                form = self.form_class(data=data)
+                self.assertIs(form.is_valid(), True)
+                self.assertEqual(form.cleaned_data, {"phone": expected})
+
+    def test_empty_region(self):
+        invalid_national_number = "8876543210"
+        form = self.form_class(data={"phone_0": "", "phone_1": invalid_national_number})
+        self.assertFalse(form.is_valid())
+        rendered_form = form.as_ul()
+        self.assertInHTML(
+            '<ul class="errorlist"><li>Enter a valid phone number '
+            "(e.g. +12125552368).</li></ul>",
+            rendered_form,
+        )
+        # Keeps national number input.
+        self.assertInHTML(
+            f'<input type="text" name="phone_1" '
+            f'value="{invalid_national_number}" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="" selected>---------</option>',
+            rendered_form,
+            count=1,
+        )
+
+    def test_empty_national_number(self):
+        form = self.form_class(data={"phone_0": "CA", "phone_1": ""})
+        self.assertFalse(form.is_valid())
+        rendered_form = form.as_ul()
+        self.assertInHTML(
+            '<ul class="errorlist"><li>The phone number entered is not valid.'
+            "</li></ul>",
+            rendered_form,
+        )
+        self.assertInHTML(
+            '<input type="text" name="phone_1" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="CA" selected>Canada +1</option>',
+            rendered_form,
+            count=1,
+        )
+
+    def test_no_region(self):
+        form = self.form_class(data={"phone_1": "654321"})
+        self.assertFalse(form.is_valid())
+        rendered_form = form.as_ul()
+        self.assertInHTML(
+            '<ul class="errorlist"><li>Enter a valid phone number '
+            "(e.g. +12125552368).</li></ul>",
+            rendered_form,
+        )
+        self.assertInHTML(
+            '<input type="text" name="phone_1" value="654321" '
+            'required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="" selected>---------</option>',
+            rendered_form,
+            count=1,
+        )
+
+    def test_no_national_number(self):
+        form = self.form_class(data={"phone_0": "CA"})
+        self.assertFalse(form.is_valid())
+        rendered_form = form.as_ul()
+        self.assertInHTML(
+            '<ul class="errorlist"><li>The phone number entered is not valid.'
+            "</li></ul>",
+            rendered_form,
+        )
+        self.assertInHTML(
+            '<input type="text" name="phone_1" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="CA" selected>Canada +1</option>',
+            rendered_form,
+            count=1,
+        )
+
+    def test_keeps_region_with_invalid_national_number(self):
+        form = self.form_class(data={"phone_0": "CA", "phone_1": "0000"})
+        self.assertFalse(form.is_valid())
+        rendered_form = str(form)
+        self.assertInHTML(
+            '<ul class="errorlist"><li>Enter a valid phone number (e.g. +12125552368).'
+            "</li></ul>",
+            rendered_form,
+        )
+        self.assertInHTML(
+            '<input type="text" name="phone_1" value="0000" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="CA" selected>Canada +1</option>',
+            rendered_form,
+            count=1,
         )
 
     def test_maxlength_not_in_select(self):
--- a/phonenumber_field/validators.py
+++ b/phonenumber_field/validators.py
@@ -1,12 +1,12 @@
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 
-from phonenumber_field.phonenumber import to_python
+from phonenumber_field.phonenumber import PhoneNumber, to_python
 
 
 def validate_international_phonenumber(value):
     phone_number = to_python(value)
-    if phone_number and not phone_number.is_valid():
+    if isinstance(phone_number, PhoneNumber) and not phone_number.is_valid():
         raise ValidationError(
             _("The phone number entered is not valid."), code="invalid_phone_number"
         )


francoisfreitag:
Early commits look good, could you move them into separate PRs so we can keep this PR focused on the change for regions sharing the same country prefix?

amateja:
Hi @francoisfreitag

I've extracted whatever was not directly and strictly linked to this fix (which wasn't that much) and I'm going to use it in separate PR. I've also reorganized commits to show as clearly as I could my journey through this case.

Regards
Andrzej

phonenumber_field/widgets.py
```diff
@@ -4,42 +4,52 @@
...
+
+
+def localized_choices(language):
+    if babel is None:
+        raise ImproperlyConfigured(
+            "The PhonePrefixSelect widget requires the babel package be installed."
+        )
+
+    choices = [("", "---------")]
+    locale_name = translation.to_locale(language=language)
```
francoisfreitag, line 34:
```suggestion
    locale_name = translation.to_locale(language)
```

phonenumber_field/widgets.py
```diff
@@ -58,22 +68,30 @@ def __init__(self, attrs=None, initial=None):
...
+
+        if isinstance(value, str):
+            value = to_python(value=value, region=region)
+
+        if isinstance(value, PhoneNumber):
+            if not value.is_valid():
+                region_code = getattr(value, "_region", region)
+                return [region_code, value.raw_input]
+            region_code = region_code_for_number(value)
+            national_number = str(value.national_number)
```
francoisfreitag, line 81: I don’t think that is correct. Some national phone numbers start with a `0`, which is significant. `national_number` is an integer, so to know how many significant `0` should be included, one must look at the `number_of_leading_zeros` and `italian_leading_zero` properties.
That’s what the `national_significant_number` does.
https://github.com/daviddrysdale/python-phonenumbers/blob/75ca34bc7c889bc3c501d00b2b1f8ec4435cb821/python/phonenumbers/phonenumberutil.py#L1690

phonenumber_field/widgets.py
```diff
@@ -58,22 +68,30 @@ def __init__(self, attrs=None, initial=None):
...
+        return [None, None]
 
     def value_from_datadict(self, data, files, name):
-        values = super().value_from_datadict(data, files, name)
-        if all(values):
-            return "%s.%s" % tuple(values)
-        return ""
+        region_code, national_number = super().value_from_datadict(data, files, name)
+
+        number = to_python(value=national_number, region=region_code)
```
francoisfreitag, line 88: The value is a positional argument.
```suggestion
        number = to_python(national_number, region=region_code)
```

tests/test_widgets.py
```diff
@@ -33,6 +51,74 @@ def test_after_translation_deactivate_all(self):
...
+
+    def test_value_from_datadict_invalid(self):
+        widget = PhoneNumberPrefixWidget()
+        data = {
+            "field_0": "US",
+            "field_1": "123",
+        }
+        value = widget.value_from_datadict(data, {}, "field")
+        self.assertEqual(value.as_e164, "+1123")
+        self.assertTrue(hasattr(value, "_region"))
```
francoisfreitag, line 82:
```suggestion
        self.assertEqual(value._region, "US")
```

tests/test_widgets.py
```diff
@@ -33,6 +51,74 @@ def test_after_translation_deactivate_all(self):
...
+
+    def test_value_from_datadict_empty(self):
+        widget = PhoneNumberPrefixWidget()
+        data = {
+            "field_0": "US",
+            "field_1": "",
+        }
+        value = widget.value_from_datadict(data, {}, "field")
+        self.assertEqual(value.raw_input, "")
+        self.assertTrue(hasattr(value, "_region"))
```
francoisfreitag, line 92:
```suggestion
        self.assertEqual(value._region, "US")
```

tests/test_widgets.py
```diff
@@ -8,22 +9,39 @@
...
+    )
+    return PhoneNumber.from_string(phone_number=e164, region=region_code)
+
+
+EXAMPLE_NUMBERS = (
+    example_number(region_code="US"),
+    example_number(region_code="CA"),
+    example_number(region_code="GB"),
+    example_number(region_code="GG"),
+    example_number(region_code="PL"),
```
francoisfreitag, line 25: Can also use an italian phone number with a significant leading 0 to cover the missed case `+390612345678`.

tests/test_widgets.py
```diff
@@ -33,6 +51,74 @@ def test_after_translation_deactivate_all(self):
...
+    @override_settings(PHONENUMBER_DEFAULT_REGION="CA")
+    def test_decompress_invalid_str(self):
+        widget = PhoneNumberPrefixWidget()
+
+        valid_number = EXAMPLE_NUMBERS[0]
+        invalid_str = f"{valid_number.as_e164}1"
+
+        invalid_number = PhoneNumber.from_string(phone_number=invalid_str)
+        self.assertFalse(invalid_number.is_valid())
+        self.assertEqual(widget.decompress(invalid_number), ["CA", invalid_str])
```
francoisfreitag, line 120: The value given to `decompress` [can be assumed to be valid](https://docs.djangoproject.com/en/4.0/ref/forms/widgets/#django.forms.MultiWidget.decompress), why is this test needed?

It may allow simplifying the `decompress` method.

phonenumber_field/widgets.py
```diff
@@ -4,42 +4,52 @@
...
-        if babel is None:
-            raise ImproperlyConfigured(
-                "The PhonePrefixSelect widget requires the babel package be installed."
-            )
-
-        choices = [("", "---------")]
         language = translation.get_language() or settings.LANGUAGE_CODE
-        locale = babel.Locale(translation.to_locale(language))
-        if not initial:
+        choices = localized_choices(language=language)
```
francoisfreitag, line 48:
```suggestion
        choices = localized_choices(language)
```

phonenumber_field/widgets.py
```diff
@@ -58,22 +68,30 @@ def __init__(self, attrs=None, initial=None):
...
-        return [None, ""]
+        region = getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
+
+        if isinstance(value, str):
+            value = to_python(value=value, region=region)
+
+        if isinstance(value, PhoneNumber):
+            if not value.is_valid():
+                region_code = getattr(value, "_region", region)
+                return [region_code, value.raw_input]
```
francoisfreitag, line 79: See comment below, the value passed to `decompress` can be assumed to be valid.
```suggestion
```

phonenumber_field/widgets.py
```diff
@@ -58,22 +68,30 @@ def __init__(self, attrs=None, initial=None):
...
-                        "+%d" % value.country_code,
-                        national_significant_number(value),
-                    ]
-            else:
-                return value.split(".")
-        return [None, ""]
+        region = getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
+
+        if isinstance(value, str):
+            value = to_python(value=value, region=region)
```
francoisfreitag, line 74:
```suggestion
            value = to_python(value, region=region)
```

tests/test_widgets.py
```diff
@@ -8,22 +9,39 @@
 )
 
 
+def example_number(region_code: str) -> PhoneNumber:
+    number = phonenumbers.example_number(region_code=region_code)
```
francoisfreitag, line 13:
```suggestion
    number = phonenumbers.example_number(region_code)
```

tests/test_widgets.py
francoisfreitag, line 25: Can pass the arguments as positional without the name. Up to you.

francoisfreitag (commented):
Great work! That’s solving some of the most active issues in the repository, coming with a nice set of tests.
Thanks a lot for the contribution. :tada:

phonenumber_field/widgets.py
```diff
@@ -4,42 +4,52 @@
...
+    for country_code, region_codes in COUNTRY_CODE_TO_REGION_CODE.items()
+    for region_code in region_codes
+}
+
+
+def localized_choices(language):
+    if babel is None:
+        raise ImproperlyConfigured(
+            "The PhonePrefixSelect widget requires the babel package be installed."
+        )
```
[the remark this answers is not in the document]
amateja: This exception in my opinion belongs to choices preparation. The `babel` library purpose is purely to translate country names. I moved choices related logic from `__init__()` method because it was a bit complex and because of testing which will follow this a bit latter if you let me contribute more. At some stage I would love to change `PhonePrefixSelect.__init__()` declaration to match it's parent class, to accept choices argument as requested in #458. If developers using our library wants to pass own choices it's up to them to provide strings translated or not. Therefore this check would not required here but it would be mandatory there.

phonenumber_field/widgets.py
amateja: I think you are right. I must have missed it.

phonenumber_field/widgets.py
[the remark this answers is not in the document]
francoisfreitag: Makes total sense, thanks for explaining! :+1:

tests/test_widgets.py
amateja: This is a really good idea. Do you mind using `+390212345678` instead because this is an example number for Italy? It would fit well to the previous test cases.

```pycon
>>> import phonenumbers
>>> it = phonenumbers.example_number('IT')
>>> phonenumbers.format_number(it, phonenumbers.PhoneNumberFormat.E164)
'+390212345678'
```

tests/test_widgets.py
francoisfreitag: Thanks for the explanation, again makes total sense (and shows serious testing has gone into this PR :100:).

I wonder if we should update the test or add another one based on an actual view to better illustrate the use case. My concern is that the current test calls directly into `decompress` and it’s not obvious to developers how `decompress` would get called in that way from user interactions.

A docstring would be an alternative, although exercising the code is a better option. I’m thinking to instantiate the form with invalid user input (inconsistent prefix-number), verify it’s reported as invalid, and verify that the form rendered with errors keeps the region, the prefix and the invalid value the user supplied.
Happy to contribute the test if that’s helpful? What do you think?

tests/test_widgets.py
francoisfreitag: Do you know if all Italian numbers have a significant leading zero? If so, then I’m fine using an example number. Otherwise, I would prefer to keep the input static so that we know the leading-zero case is covered, even if `phonenumbers` changes their example number for the italian region.

francoisfreitag: Haha no worries. Either way works.
The main rationale for me is:
1. Supplying positional argument positionally. The method signature has been made to accept the argument positionally, there might be a logic in their order. It also helps identifying their nature (e.g. order matters)
2. Avoid repeating the argument name (e.g. `can_view(user=user)`).
3. Slightly better performance https://peps.python.org/pep-0570/#performance
4. It helps with refactoring and grepping through the code, with more uniform call patterns.

amateja: According to [ITU-T](https://www.itu.int/dms_pub/itu-t/oth/02/02/T020200006B0001PDFE.pdf) all Italian geographic numbers starts with zero and `phonenumbers.example_number` returns a valid fixed-line number for the specified region. So this is confirmed.

amateja: This is covered with 2a9a1d9.

phonenumber_field/widgets.py
amateja: This is covered with 2a9a1d9.

tests/test_widgets.py
francoisfreitag: I may have missed something, but the nomadic communications numbers (I’m guessing mobile) start with 55, not 0.

Constructing the phone number ourselves is preferable, because we do not control what example number is chosen when only specifying the "IT" region. WDYT?

amateja: Basically `example_number` is just a w wrapper over `example_number_for_type` to always return `PhoneNumberType.FIXED_LINE`. Those numbers always start with 0 in Italy.

```pycon
>>> from phonenumbers import PhoneNumberFormat, PhoneNumberType, example_number_for_type, format_number
>>> 
>>> example = example_number_for_type('IT', PhoneNumberType.FIXED_LINE)
>>> print(format_number(example, PhoneNumberFormat.E164))
+390212345678
```

What you mentioned is `PhoneNumberType.VOIP` and it starts with 55 indeed.

```pycon
>>> from phonenumbers import PhoneNumberFormat, PhoneNumberType, example_number_for_type, format_number
>>> 
>>> example = example_number_for_type('IT', PhoneNumberType.VOIP)
>>> print(format_number(example, PhoneNumberFormat.E164))
+395512345678
```
Mobile numbers on the other hand are type `PhoneNumberType.MOBILE` and starts with 3.

```pycon
>>> from phonenumbers import PhoneNumberFormat, PhoneNumberType, example_number_for_type, format_number
>>> 
>>> example = example_number_for_type('IT', PhoneNumberType.MOBILE)
>>> print(format_number(example, PhoneNumberFormat.E164))
+393123456789
```

As long as Autorità per le Garanzie nelle Comunicazion keeps this format there should be no changes and there always will be leading zero. If they don't our test case would also fail because our static example number would no longer be correct.

I understand your concern and I think this is safe to transfer this responsibility to our dependency because fixed line numbers in Italy share this feature and we always get correct example number.

francoisfreitag: Ok. The default being `FIXED_LINE` works for me.

tests/test_widgets.py
amateja: I wouldn't test it too much and on too high level because developers have some flexibility and they don't need to use `PhoneNumberPrefixWidget` straightforwardly on the one hand and we might end up testing Django internals on the other hand. That's why I focused myself on testing those parts which are overridden in the child class.

One level above a widget we have a form field which is a part of model from

francoisfreitag: > It goes like this:
> 
>     1. user selects correct region and gives correct number but not in contexts of selected region
> 
>     2. `PhoneNumberPrefixWidget.value_from_datadict` is called and we get a `PhoneNumber` instance with invalid value
> 
>     3. `formfields.PhoneNumberField` is instantiated and we get a form error
> 
>     4. to render e.g. admin page with error message and request to correct given input `PhoneNumberPrefixWidget.decompress` is called with invalid `PhoneNumber` instance

I spent some time today trying to reproduce these STR and failed. Here’s my test project 
[test.tar.gz](https://github.com/stefanfoulis/django-phonenumber-field/files/8398566/test.tar.gz).

Here’s the output of pip-list:
```
Package                  Version                        Editable project location
------------------------ ------------------------------ -------------------------------------------
asgiref                  3.5.0
Babel                    2.9.1
black                    22.3.0
click                    8.1.2
Django                   4.0.3
django-phonenumber-field 6.1.1.dev12+g2a9a1d9.d20220401 /home/freitafr/dev/django-phonenumber-field
flake8                   4.0.1
isort                    5.10.1
mccabe                   0.6.1
mypy-extensions          0.4.3
pathspec                 0.9.0
phonenumberslite         8.12.46
pip                      22.0.4
platformdirs             2.5.1
pycodestyle              2.8.0
pyflakes                 2.4.0
pytz                     2022.1
setuptools               58.1.0
sqlparse                 0.4.2
tomli                    2.0.1
```

Here’s the patch I applied locally expecting a failure:
```diff
diff --git a/phonenumber_field/widgets.py b/phonenumber_field/widgets.py
index cd76de1..5dc8038 100644
--- a/phonenumber_field/widgets.py
+++ b/phonenumber_field/widgets.py
@@ -75,9 +75,9 @@ class PhoneNumberPrefixWidget(MultiWidget):
             value = to_python(value, region=region)
 
         if isinstance(value, PhoneNumber):
-            if not value.is_valid():
-                region_code = getattr(value, "_region", region)
-                return [region_code, value.raw_input]
+            # if not value.is_valid():
+            #     region_code = getattr(value, "_region", region)
+            #     return [region_code, value.raw_input]
             region_code = region_code_for_number(value)
             national_number = national_significant_number(value)
             return [region_code, national_number]
@@ -90,8 +90,6 @@ class PhoneNumberPrefixWidget(MultiWidget):
         if isinstance(number, str):
             number = PhoneNumber()
             number.raw_input = national_number
-        if not number.is_valid():
-            setattr(number, "_region", region_code)
         return number
 ```

Steps I took:
1. Create a user `./manage.py createsuperuser` 
2. Login as that user
3. http://127.0.0.1:8000/admin/app/mymodel/add/
4. Choose France as the country and enter invalid national number: `8876543210`

---


FWIW, the following test also passes with the patch:
```python
    @override_settings(PHONENUMBER_DEFAULT_REGION="CA")
    def test_decompress_invalid_str(self):
        class TestForm(forms.Form):
            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)

        invalid_national_number = "8876543210"
        form = TestForm(data={"phone_0": "FR", "phone_1": invalid_national_number})
        self.assertIs(form.is_valid(), False)
        rendered_form = str(form)
        # Retain selected region.
        self.assertInHTML(
            """
            <ul class="errorlist">
            <li>
                Enter a valid phone number (e.g. (506) 234-5678) or a number
                with an international call prefix.
            </li>
            </ul>
            """,
            rendered_form,
        )
        self.assertInHTML(
            '<option value="FR" selected>France +33</option>',
            rendered_form,
            count=1,
        )
        self.assertInHTML(
            f'<input type="text" name="phone_1" value="{invalid_national_number}" required id="id_phone_1">',
            rendered_form,
            count=1,
        )
```

---

I am still unclear on when decompress is passed invalid data?

amateja: I just used your project and latest state of this PR (commmit 2a9a1d9). I added in [widgets.py](phonenumber_field/widgets.py) a new raising exception line after line 78 like this:

```python
        if isinstance(value, PhoneNumber):
            if not value.is_valid():
                raise ValueError(value)
                region_code = getattr(value, "_region", region)
                return [region_code, value.raw_input]
```
instead of this
```python
        if isinstance(value, PhoneNumber):
            if not value.is_valid():
                region_code = getattr(value, "_region", region)
                return [region_code, value.raw_input]
```

Then I navigated to `/admin/app/mymodel/add/` URL and selected `France +33` and I hit `SAVE`. Please note empty national number and voilà! I've got pretty nasty exception as expected.

This also works for let say `France +33` and `000000`.

amateja: And your patch rolls back changes which help for #461.

francoisfreitag: What I am looking for is noticeable issues from a user perspective, to write tests that map reasonably closely to usage instead of being coupled to the implementation. That’s my main concern with the `test_decompress`. That’s why I removed some bits of code and tried using the app, expecting a failure. I was not suggesting to apply my patch to this PR, sorry for the misunderstanding.
Adding an exception to show the code path is hit does not help in building a test case that’s representative of the expected behavior.

With the patch above, the invalid form is displayed correctly, with the selected country code and the national number. That’s why I was asking for the kind of failures the STR were addressing.

> Then I navigated to /admin/app/mymodel/add/ URL and selected France +33 and I hit SAVE. Please note empty national number [...]

That’s the kind of info I was looking for! In trying out more today, I found a couple issues captured in the following tests (only two tests show incorrect behavior from the system, the others are provided because I think they should be added to the test suite):
```diff
diff --git a/tests/test_widgets.py b/tests/test_widgets.py
index d632e71..fa1804e 100644
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -1,8 +1,10 @@
 import phonenumbers
+from django import forms
 from django.test import SimpleTestCase, override_settings
 from django.utils import translation
 
 from phonenumber_field.phonenumber import PhoneNumber
+from phonenumber_field import formfields
 from phonenumber_field.widgets import (
     PhoneNumberInternationalFallbackWidget,
     PhoneNumberPrefixWidget,
@@ -107,16 +109,103 @@ class PhoneNumberPrefixWidgetTest(SimpleTestCase):
             with self.subTest(value):
                 self.assertEqual(widget.decompress(value), expected)
 
-    @override_settings(PHONENUMBER_DEFAULT_REGION="CA")
-    def test_decompress_invalid_str(self):
-        widget = PhoneNumberPrefixWidget()
+    def test_empty_region(self):
+        class TestForm(forms.Form):
+            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)
+
+        invalid_national_number = "8876543210"
+        form = TestForm(data={"phone_0": "", "phone_1": invalid_national_number})
+        self.assertIs(form.is_valid(), False)
+        rendered_form = str(form)
+        self.assertInHTML(
+            """
+            <ul class="errorlist"><li>Enter a valid phone number (e.g. +12125552368).</li></ul>
+            """,
+            rendered_form,
+        )
+        # Keeps national number input.
+        self.assertInHTML(
+            f'<input type="text" name="phone_1" value="{invalid_national_number}" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="" selected>---------</option>',
+            rendered_form,
+            count=1,
+        )
+
+    def test_empty_national_number(self):
+        class TestForm(forms.Form):
+            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)
+
+        form = TestForm(data={"phone_0": "CA", "phone_1": ""})
+        self.assertIs(form.is_valid(), False)
+        rendered_form = str(form)
+        self.assertInHTML(
+            """
+            <ul class="errorlist"><li>Enter a valid phone number (e.g. +12125552368).</li></ul>
+            """,
+            rendered_form,
+        )
+        self.assertInHTML(
+            f'<input type="text" name="phone_1" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="CA" selected>Canada +1</option>',
+            rendered_form,
+            count=1,
+        )
 
-        valid_number = EXAMPLE_NUMBERS[0]
-        invalid_str = f"{valid_number.as_e164}1"
+    def test_no_region(self):
+        class TestForm(forms.Form):
+            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)
+
+        form = TestForm(data={"phone_1": "654321"})
+        self.assertIs(form.is_valid(), False)
+        rendered_form = str(form)
+        self.assertInHTML(
+            """
+            <ul class="errorlist"><li>Enter a valid phone number (e.g. +12125552368).</li></ul>
+            """,
+            rendered_form,
+        )
+        self.assertInHTML(
+            f'<input type="text" name="phone_1" value="654321" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="" selected>---------</option>',
+            rendered_form,
+            count=1,
+        )
 
-        invalid_number = PhoneNumber.from_string(invalid_str)
-        self.assertFalse(invalid_number.is_valid())
-        self.assertEqual(widget.decompress(invalid_number), ["CA", invalid_str])
+    def test_no_national_number(self):
+        class TestForm(forms.Form):
+            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)
+
+        form = TestForm(data={"phone_0": "CA"})
+        self.assertIs(form.is_valid(), False)
+        rendered_form = str(form)
+        self.assertInHTML(
+            """
+            <ul class="errorlist"><li>Enter a valid phone number (e.g. +12125552368).</li></ul>
+            """,
+            rendered_form,
+        )
+        self.assertInHTML(
+            f'<input type="text" name="phone_1" value="" required id="id_phone_1">',
+            rendered_form,
+            count=1,
+        )
+        self.assertInHTML(
+            '<option value="CA" selected>Canada +1</option>',
+            rendered_form,
+            count=1,
+        )
 
 
 class PhoneNumberInternationalFallbackWidgetTest(SimpleTestCase):
```

```
======================================================================
ERROR: test_no_national_number (tests.test_widgets.PhoneNumberPrefixWidgetTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/freitafr/dev/django-phonenumber-field/tests/test_widgets.py", line 191, in test_no_national_number
    self.assertIs(form.is_valid(), False)
  File "/home/freitafr/dev/django-phonenumber-field/.tox/py310/lib/python3.10/site-packages/django/forms/forms.py", line 205, in is_valid
    return self.is_bound and not self.errors
  File "/home/freitafr/dev/django-phonenumber-field/.tox/py310/lib/python3.10/site-packages/django/forms/forms.py", line 200, in errors
    self.full_clean()
  File "/home/freitafr/dev/django-phonenumber-field/.tox/py310/lib/python3.10/site-packages/django/forms/forms.py", line 433, in full_clean
    self._clean_fields()
  File "/home/freitafr/dev/django-phonenumber-field/.tox/py310/lib/python3.10/site-packages/django/forms/forms.py", line 440, in _clean_fields
    value = bf.initial if field.disabled else bf.data
  File "/home/freitafr/dev/django-phonenumber-field/.tox/py310/lib/python3.10/site-packages/django/forms/boundfield.py", line 127, in data
    return self.form._widget_data_value(self.field.widget, self.html_name)
  File "/home/freitafr/dev/django-phonenumber-field/.tox/py310/lib/python3.10/site-packages/django/forms/forms.py", line 224, in _widget_data_value
    return widget.value_from_datadict(self.data, self.files, html_name)
  File "/home/freitafr/dev/django-phonenumber-field/phonenumber_field/widgets.py", line 93, in value_from_datadict
    if not number.is_valid():
AttributeError: 'NoneType' object has no attribute 'is_valid'

======================================================================
FAIL: test_empty_national_number (tests.test_widgets.PhoneNumberPrefixWidgetTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/freitafr/dev/django-phonenumber-field/tests/test_widgets.py", line 143, in test_empty_national_number
    self.assertIs(form.is_valid(), False)
AssertionError: True is not False
```

---

> And your patch rolls back changes which help for https://github.com/stefanfoulis/django-phonenumber-field/issues/461.

Yes, because I was also trying to catch the incorrect behavior without the private `_region` attr. The current tests rely on an implementation detail, which is not as helpful as verifying the expected behavior. The presence of the attribute does not indicate that it is used, nor its purpose.
Here’s an example test verifying the behavior.
```python
    def test_keeps_region_with_invalid_national_number(self):
        class TestForm(forms.Form):
            phone = formfields.PhoneNumberField(widget=PhoneNumberPrefixWidget)

        form = TestForm(data={"phone_0": "CA", "phone_1": "0000"})
        self.assertIs(form.is_valid(), False)
        rendered_form = str(form)
        self.assertInHTML(
            """
            <ul class="errorlist"><li>Enter a valid phone number (e.g. +12125552368).</li></ul>
            """,
            rendered_form,
        )
        self.assertInHTML(
            f'<input type="text" name="phone_1" value="0000" required id="id_phone_1">',
            rendered_form,
            count=1,
        )
        self.assertInHTML(
            '<option value="CA" selected>Canada +1</option>',
            rendered_form,
            count=1,
        )
```

amateja: Please find your patch slightly updated as a part of my recent commits. I really appreciate those examples you gave because now it's clearer that you are aiming for functional tests over unit tests. I let myself to improve few lines here and there but all kudos goes to you.

I also discovered that in very edge case scenario which is rather not probably `None` value can be returned in `value_from_datadict` causing the exception you referred to. This is fixed in a9ed219. There is one more issue with validation where invalid `PhoneNumber` instance with `raw_input` property set to empty string is evaluated to `False` in `if` clause in: https://github.com/stefanfoulis/django-phonenumber-field/blob/022dccc3caffe3aba0ea8388fb3d4a2201f2f0fb/phonenumber_field/validators.py#L7-L12
This is also fixed in c05200a.

I hope this exhaust all uncertainties in this thread. I'm really looking forward to your feedback on recent improvements. I also hope this makes this PR complete.

francoisfreitag (commented):
Thanks for the rework, this is definitely close to landing. It’s a valuable contribution and the project will definitely be improved by it, thanks a lot for taking the time to upstream it!
The discussion on the patch is only meant to ensure its maintainability, trying hard to avoid mistakes. Your work is definitely appreciated and most welcome :+1:.

Remaining open points:
- How can an `str` value be passed to `decompress()`?
- Please write the new tests at a more functional level, that better explain the use case than a unit test calling into Django multi widget methods. It’s also much easier to call Django multi widget method in a non-representative than when manipulating the form.

**Edit:** Removed the `str` guard from decompress and corresponding test.

J4bbi:
Many thanks @amateja for this quality PR. It seems to fix exactly the issue I am having (Guernsey and UK sharing prefixes).

If possible, could you please respond to @francoisfreitag 's issues so that this might be merged?

If @amateja is too busy. Can we consider either pushing to this PR (not sure if that's possible) or re-submitting one with code you are 100% okay with @francoisfreitag ?

My aim here is not to show anyone disrespect in any way, just to improve the library. Otherwise I will have to apply @amateja 's fix directly to the library.

francoisfreitag:
Ideally, @amateja can take this work to the finish line. His contribution is great, and I’m eager to integrate it. The question blocking me from merging currently is “How can an `str` value be passed to `decompress()`” (and related https://github.com/stefanfoulis/django-phonenumber-field/pull/493#discussion_r855203242)?
A test is the best way to demonstrate the use case, but even steps to reproduce or details about how that can happen are sufficient. I am happy to write the corresponding test and apply the edits suggested in past reviews.

Of course, if others interested in this PR understand the reason for [that guard](https://github.com/stefanfoulis/django-phonenumber-field/pull/493/files#diff-a0bd90f5b44b3524a51b2d693276af85370b54a0cdaacdf42eec6d8fdadea307R74), please share it.

francoisfreitag:
Without response, I have gone ahead and applied the last cleanups to this patch.
Integrating and releasing it soon.

Many thanks to @amateja, it’s a great contribution to this project! :star2:

francoisfreitag (approved):

PR #495, opened 2022-03-14
expanding tests coverage

amateja:
Reviewing recent issues and preparing recent PR I've written some additional tests which expands tests coverage. Hope you'll find them useful.

### DIFF
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -1,11 +1,28 @@
+from django.core.exceptions import ImproperlyConfigured
 from django.test import SimpleTestCase, override_settings
 from django.utils import translation
 
+from phonenumber_field import widgets
 from phonenumber_field.phonenumber import PhoneNumber
 from phonenumber_field.widgets import (
     PhoneNumberInternationalFallbackWidget,
     PhoneNumberPrefixWidget,
 )
+
+
+class PhonePrefixSelectTest(SimpleTestCase):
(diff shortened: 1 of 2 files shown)


tests/test_widgets.py
```diff
@@ -8,22 +9,33 @@
 )
 
 
+class PhonePrefixSelectTest(SimpleTestCase):
+    def test_without_babel(self):
+        import phonenumber_field
+
+        babel_module = phonenumber_field.widgets.babel
+        phonenumber_field.widgets.babel = None
```
francoisfreitag, line 17: Please make sure the cleanup is always executed, e.g. with https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup.

francoisfreitag (approved):
Otherwise LGTM, thanks!

tests/test_widgets.py
amateja: Good point. I think using `setUp` and `tearDown` should be enough. The only difference between `tearDown` and registering a function with `addCleanup` is that if won't be fired if `setUp` fails whilst cleanup method is fired always.

francoisfreitag:
Thanks for the update. Pushed cosmetic edits. If they look good to you, I’ll merge this.

amateja:
I totally don't mind rephrasing commit messages or splitting and joining my commits. Thanks for your help.

PR #498, opened 2022-03-20
better syntax highlighting

bashu:


bashu:
supported by pypi.org too, so colours will be a way nicer

amateja (approved):

PR #515, opened 2022-08-16
Drop support for end-of-life Django 2.2

francoisfreitag:

PR #520, opened 2022-08-27
Render input value with PhoneNumberPrefixWidget on form errors

### LINKED ISSUE #518: PhoneNumberPrefixWidget - User input is cleared (on form errors)
When re-rendering form with errors, the user input for phone number is cleared. **But only when the default region was selected**. When user selected a different region than the one specified in PHONENUMBER_DEFAULT_REGION, his input persists. 
Entered phone number is not present even if it was valid. 

At first I was certain the error lied somewhere in my code, but after building a bare bone form and view I concluded it has to be a bug. FYI I have just upgraded to django 4.1 (from 3.2) and all other packages.

I'm currently using:
- Python 3.10
- Django 4.1
- django-phonenumber-field 6.3.0 (tried also dev version, which solves ignoring required=True on field at least)

My bare-bone example:

Settings.py
`PHONENUMBER_DEFAULT_REGION = 'SI'`

Forms.py
```
from phonenumber_field.formfields import PhoneNumberField
from phonenumber_field.widgets import PhoneNumberPrefixWidget

class TestForm(forms.Form):
    name = forms.CharField(max_length=200, min_length=3) # Just here for triggering an invalid form
    tel = PhoneNumberField(required=False, widget=PhoneNumberPrefixWidget())
```

Views.py
```
from .forms import TestForm
class TestFormView(FormView):
    form_class= TestForm
    template_name="test_form.html"
```

Html
```{% extends 'base.html' %}

{% block content %}
<div class="container-fluid">
    <form method="POST" novalidate>
        {% csrf_token %}
        {{ form }}
        <input type="submit" name="submit">
    </form>    
</div>
{% endblock %}
```

francoisfreitag:
The phone number input field was incorrectly cleared when:
- a form has an error on another field, and
- the phone number field is displayed with PhoneNumberPrefixWidget, and
- the phone number region is the default region.

That behavior was caused by 005769cf39323e5b23710783f45befb546672cd6,
which incorrectly used `prepare_value()` to transform a PhoneNumber
instance to an str (the phone number in national format). By doing so,
information was lost and the decompress() method of
PhoneNumberPrefixWidget unexpectedly received an str.

The phone number value is now always a PhoneNumber instance. It’s up to
the widget to format that value.
To retain the behavior of showing national phone numbers in the national
format by default, the formfields.PhoneNumberField widget is now
PhoneNumberInternationalFallbackWidget.

PhoneNumberInternationalFallbackWidget behavior was updated to match the
previous implementation of prepare_value: the number is displayed in
national format if its country code matches the country code of the
configured region.
It now represents its value as a PhoneNumber, a richer type than `str`,
which facilitates the custom display logic.

Fixes #518

### DIFF
--- a/phonenumber_field/formfields.py
+++ b/phonenumber_field/formfields.py
@@ -6,19 +6,27 @@ from django.forms.fields import CharField
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
 
-from phonenumber_field.phonenumber import PhoneNumber, to_python, validate_region
+from phonenumber_field.phonenumber import to_python, validate_region
 from phonenumber_field.validators import validate_international_phonenumber
+from phonenumber_field.widgets import PhoneNumberInternationalFallbackWidget
 
 
 class PhoneNumberField(CharField):
     default_validators = [validate_international_phonenumber]
+    widget = PhoneNumberInternationalFallbackWidget
 
-    def __init__(self, *args, region=None, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.widget.input_type = "tel"
-
+    def __init__(self, *args, region=None, widget=None, **kwargs):
         validate_region(region)
         self.region = region or getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
+
+        if widget is None:
+            try:
+                widget = self.widget(region=self.region)
+            except TypeError:
+                widget = self.widget()
+
+        super().__init__(*args, widget=widget, **kwargs)
+        self.widget.input_type = "tel"
 
         if "invalid" not in self.error_messages:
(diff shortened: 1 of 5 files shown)


francoisfreitag:
We should bump the release to 7.0.0 with this, due to the change of the `formfields.PhoneNumberField` default widget to `PhoneNumberInternationalWidget`, which may introduce some subtle backward incompatibilities.

---
Django 2.2 failures can be ignored, the version is dead. See https://github.com/stefanfoulis/django-phonenumber-field/pull/515.

francoisfreitag:
CC @amateja, that explains why `decompress()` used to receive an `str` instance.

stefanfoulis (approved):

PR #521, opened 2022-08-27
Set PhoneNumberInternationalFallbackWidget input_type to tel

francoisfreitag:


phonenumber_field/widgets.py
```diff
@@ -103,6 +103,8 @@ class PhoneNumberInternationalFallbackWidget(TextInput):
     an international number will fall back to international format
     """
 
+    input_type = "tel"
```
stefanfoulis, line 106: I like this change, but I wonder if it will break some peoples code/styling. I think it needs to be mentioned in release notes and maybe even with a bigger version bump :thinking: .

stefanfoulis (approved):

phonenumber_field/widgets.py
francoisfreitag: That’s a good idea. I’ll add it to the 7.0.0 with #520 (which may also introduce a small BC breakage).

PR #529, opened 2022-08-31
Evolve PhoneNumberInternationalWidget to RegionalPhoneNumberWidget

francoisfreitag:
Instead of hardcoding the INTERNATIONAL format, follow the
PHONENUMBER_DEFAULT_FORMAT preference (defaulting to E164).

The PhoneNumberInternationalWidget is now deprecated and will be removed
in the next major version. Use PhoneNumberWithFallbackWidget with
setting PHONENUMBER_DEFAULT_FORMAT="INTERNATIONAL" instead.


francoisfreitag:
I would like to include this change in the upcoming 7.0.0, so that we keep the breaking changes together.

stefanfoulis (approved):

PR #530, opened 2022-09-01
Test serialization of phone numbers

francoisfreitag:


francoisfreitag:
Django main tests will work again when https://github.com/encode/django-rest-framework/commit/ad282da97cf7b23c50a8fa7b7c5cad68c1deedc3 is released.

PR #543, opened 2022-11-23
Allow multiple DRF is_valid calls to succeed

### LINKED ISSUE #542: Cannot re-validate validated_data in DRF
This is a bit of a specific issue with writable nested serializers and the changed behavior in #523. Since 6.4.0 many of my phone number fields now fail validation.

Here is a very contrived serializer makeup that illustrates the problem:

```python
class NestedSerializer(Serializer):
    primary_phone_number = FixedPhoneNumberField(required=False, allow_null=True, allow_blank=True)

    def create(self, validated_data):
        # do lots of business logic
        return validated_data


class ParentSerializer(Serializer):
    nested = NestedSerializer(write_only=True)
    anotherfield = CharField()

    def create(self, validated_data):
        nested_instance = NestedSerializer(data=validated_data.pop("nested"))
        nested_instance.is_valid(raise_exception=True)
        nested_instance.save()
        return validated_data
```

Here's an example of using these serializers:

```python
class PhoneNumberRegressionTestCase(TestCase):

    def test_serializers(self):
        parent = ParentSerializer(data={"nested": {"primary_phone_number": "+15126121234"}, "anotherfield": "some data"})
        parent.is_valid(raise_exception=True)
        parent.save()
```

When I run this code, I get back a `ValidationError` on `nested.primary_phone_number`:

```
  File "/Users/phillip/demo-proj/tests/test_serializers.py", line 319, in create
    nested_instance.is_valid(raise_exception=True)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/demo-proj-cpwchxt0-py3.10/lib/python3.10/site-packages/rest_framework/serializers.py", line 235, in is_valid
    raise ValidationError(self.errors)
rest_framework.exceptions.ValidationError: {'primary_phone_number': [ErrorDetail(string='Enter a valid phone number.', code='invalid')]}
```

Digging into this, this is because the DRF `CharField` [only accepts a `str`, `int`, `float` or `bool`](https://github.com/encode/django-rest-framework/blob/ebde56b932e638c83d3a164ca89b618c520d0ca2/rest_framework/fields.py#L756-L757). In the change in #523, we unconditionally pass the `data` field to the super method. However, in this piece of code in the `create()`:

```python
    def create(self, validated_data):
        nested_instance = NestedSerializer(data=validated_data.pop("nested"))
```

The `primary_phone_number` entry is already a `PhoneNumber` because of the implementation of `to_internal_value`. Initializing the `NestedSerializer` again with the `validated_data` runs through validation again (with the `is_valid` call) and the new call to `super().to_initial_value()` fails.

I thought perhaps the fix would be to use `self.initial_data` instead of `validated_data` which has the original string instance, but [this goes against the DRF docs](https://www.django-rest-framework.org/api-guide/relations/#writable-nested-serializers) of how to use writable nested serializers.

Here is a workaround that fixes the issue and is as flexible as versions < 6.4.0:

```python
from phonenumber_field.phonenumber import to_python
from rest_framework.exceptions import ValidationError

class FixedPhoneNumberField(PhoneNumberField):

    def to_internal_value(self, data):

        if isinstance(data, PhoneNumber):
            phone_number = data
        else:
            str_value = super().to_internal_value(data)
            phone_number = to_python(str_value, region=self.region)

        if phone_number and not phone_number.is_valid():
            raise ValidationError(self.error_messages["invalid"])
        return phone_number
```

phillipuniverse:
Fixes #542

### DIFF
--- a/phonenumber_field/serializerfields.py
+++ b/phonenumber_field/serializerfields.py
@@ -19,8 +19,12 @@ class PhoneNumberField(serializers.CharField):
         self.region = region or getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
 
     def to_internal_value(self, data):
-        str_value = super().to_internal_value(data)
-        phone_number = to_python(str_value, region=self.region)
+        if isinstance(data, PhoneNumber):
+            phone_number = data
+        else:
+            str_value = super().to_internal_value(data)
+            phone_number = to_python(str_value, region=self.region)
+
         if phone_number and not phone_number.is_valid():
             raise ValidationError(self.error_messages["invalid"])
         return phone_number


phillipuniverse:
Happy to go through the process of adding tests, etc but I wanted to make sure this is a reasonable change!

francoisfreitag:
As noted in the issue, the change is reasonable and fixing a bug in the library (when comparing to DRF builtin serializer fields. Thank you for reporting (and fixing!) it.
Feel free to add a test. When the PR is ready, I’ll merge your work and make a release.

phillipuniverse:
@francoisfreitag test added and missing import fixed!

francoisfreitag:
Great, thank you!

PR #547, opened 2023-02-26
Add common virtualenv locations to flake8 exclude and .gitignore

### LINKED ISSUE #546: flake8 doesn't exclude common virtualenv locations
It's common practice (especially with IDE's) when developing on a project to put a virtualenv in either .venv or venv at the git root, for development.

Currently this is an issue for the flake8 tests, since, when called without arguments it now flake8 checks all the .py files in your virtual environment, which produces a cavalcade of errors.  :cry:

Either flake8 should be called with arguments, or it's exclude patterns should be extended to include these paths.

blaedd:
A minor quality of life improvement for (potential) contributors, this stops flake8 generating a barrage of errors when run, if you have a virtualenv in one of the common in-tree locations (.venv/, venv/ .env/).

Adds these paths to both flake8 extend-ignore in setup.cfg and .gitignore

Fixes #546

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,8 @@ share/
 .coverage*
 # Written by setuptools_scm.
 phonenumber_field/version.py
+# python virtualenv
+.venv/
+venv/
+# python virtualenv or environment file (potential secrets)
+.env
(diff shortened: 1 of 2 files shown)


francoisfreitag (approved):
Sure, thanks!

PR #557, opened 2023-05-13
Switch to pyproject.toml

francoisfreitag:
Starting with [PEP 621](https://peps.python.org/pep-0621/), the Python community selected pyproject.toml as a standard way of specifying project metadata. Setuptools has adopted this standard and will use the information contained in this file as an input in the build process.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,53 @@
+[project]
+name = "django-phonenumber-field"
+requires-python = ">=3.7"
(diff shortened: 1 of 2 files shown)


francoisfreitag:
I dropped the `platforms` directive from `pyproject.toml`, it caused an error on `python -m build`:
```
ValueError: invalid pyproject.toml config: `project`.
configuration error: `project` must not contain {'platforms'} properties
```

Issue #558, opened 2023-05-14
`.values` and `.values_list` return `str` instead of `PhoneNumber` object


sssanikhani:
I have this model:
```  py
class A(models.Model):
    phone = PhoneNumberField()
```

When I want to get `values_list` of some objects, expected return type is list of `PhoneNumber`s but it returns list of `str`s:
``` py
list(A.objects.all().values_list('phone', flat=True))
```

Output
``` sh
['+989120000000', '+989121111111', '+989122222222']
```

PR #559, opened 2023-05-15
Add a DB converter for the model field

francoisfreitag:
#558

### DIFF
--- a/phonenumber_field/modelfields.py
+++ b/phonenumber_field/modelfields.py
@@ -102,6 +102,9 @@ class PhoneNumberField(models.CharField):
 
         return super().get_prep_value(value)
 
+    def from_db_value(self, value, expression, connection):
+        return to_python(value)
+
     def contribute_to_class(self, cls, name, *args, **kwargs):
         super().contribute_to_class(cls, name, *args, **kwargs)
         setattr(cls, self.name, self.descriptor_class(self))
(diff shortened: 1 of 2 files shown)


kalekseev:
This is a breaking change, would be better to release it in v8...

francoisfreitag:
Oops, that’s right. Sorry for the trouble! I’ll update the release note text to highlight that.

chris-erickson:
How would we get the old behavior back?

francoisfreitag:
Depending on what you expect, I would say either:
- call `str()` on the `PhoneNumber` object (or the property for a specific format, such as [`.as_international`](https://django-phonenumber-field.readthedocs.io/en/latest/reference.html#phonenumber_field.phonenumber.PhoneNumber.as_international), [`.as_e164`](https://django-phonenumber-field.readthedocs.io/en/latest/reference.html#phonenumber_field.phonenumber.PhoneNumber.as_e164), ...)
- subclass the `phonenumber_field.modelfields.PhoneNumberField` and implement `to_python()` as     
  ```python
    def to_python(self, value):
        return value
   ```

Neumsy:
> * subclass the `phonenumber_field.modelfields.PhoneNumberField` and implement `to_python()` as
>   ```python
>     def to_python(self, value):
>         return value
>   ```

When using this method, shouldn't the suggestion be to override the from_db_value method?  The calling to_python method is not part of the class inheritance, so overriding it in the class doesn't produce the desired outcome.

```python
    def from_db_value(self, value, expression, connection):
        return str(value)
```

PR #579, opened 2023-10-06
Refactor: Phone Number Serializer Field

keshavmohta09:
Add `get_str_value` and `split_representation` support

- `get_str_value`:  boolean value to get string type of phone number in validated data when needed
- `split_representation`: boolean value to represent country code and number separately

### DIFF
--- a/phonenumber_field/serializerfields.py
+++ b/phonenumber_field/serializerfields.py
@@ -9,14 +9,29 @@ from phonenumber_field.phonenumber import PhoneNumber, to_python, validate_regio
 class PhoneNumberField(serializers.CharField):
     default_error_messages = {"invalid": _("Enter a valid phone number.")}
 
-    def __init__(self, *args, region=None, **kwargs):
+    def __init__(
+        self,
+        *args,
+        region=None,
+        get_str_value=False,
+        split_representation=False,
+        **kwargs,
+    ):
         """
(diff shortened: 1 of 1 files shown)


francoisfreitag:
Hi, thanks for submitting a PR!

Could you explain why these changes should be made?

Currently, I am thinking of rejecting both:

- `get_str_value`, from [django REST framework](https://www.django-rest-framework.org/api-guide/fields/#custom-fields) docs:

  > The `.to_internal_value()` method is called to restore a primitive datatype into its internal python representation.

  The internal Python representation of a phone number should be a `PhoneNumber` instance.

- `to_representation`, I like the fact that the representation can be fed back to the serializer. With this proposal, we would need to change `to_internal_value()` to handle the split representation. It seems easy to subclass `PhoneNumberSerializer` to add your own `to_representation`, if the custom implementation does not fit your use case. But I’ld like to hear about your use case :smiley:

keshavmohta09:
Hello, thanks for viewing the changes.

I agree that the phone number instance should be returned from the to_internal_value method, but in many cases, we need its string value, so I have added a flag.

In the case of split_representation, in most cases the country code and phone number need to be sent separately to the frontend, and displayed separately, that's why I have added.

keshavmohta09:
Okay, thankyou for guiding me. I am closing this PR.

PR #583, opened 2023-11-15
Document phone number extension handling

### LINKED ISSUE #581: Extension information is lost during serialization
Firstly, thanks for an awesome library!

It looks to me like the extension information is lost when using Django's built in serialization. We only use this for loading our test database, but it was surprising when the extension's didn't survive reloading the data.

After a little bit of investigation, I think fixing this should be pretty straight forwards, we would just need to implement `value_to_string()` for the field with one of the lossless serialization formats and then make sure the the `to_python()` function can handle it properly.

https://docs.djangoproject.com/en/4.2/howto/custom-model-fields/#converting-field-data-for-serialization

If the contribution is welcome, I might be interested in adding this fix. Thanks again!

francoisfreitag:
Closes #581

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -44,4 +44,5 @@ your ``settings.py`` file :external+django:setting:`INSTALLED_APPS`:
    :maxdepth: 2
    :caption: Contents:
 
+   phonenumbers
    reference
--- a/docs/phonenumbers.rst
+++ b/docs/phonenumbers.rst
@@ -1,0 +1,31 @@
+Handling phone numbers
+======================
+
+`Google’s libphonenumber “Falsehoods Programmers Believe About Phone Numbers”
+<https://github.com/google/libphonenumber/blob/master/FALSEHOODS.md>`_ are
+worth keeping in mind, especially since this library relies heavily on
+`libphonenumber <https://github.com/google/libphonenumber/>`_.
+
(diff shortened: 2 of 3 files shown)


francoisfreitag:
@stefanfoulis do you remember why E.164 is the default over E.123 (INTERNATIONAL)?

stefanfoulis:
> @stefanfoulis do you remember why E.164 is the default over E.123 (INTERNATIONAL)?
I don't know. :-/

francoisfreitag:
I see. It might be a good idea to transition to the `INTERNATIONAL` format by default. I’ll start the deprecation in the upcoming weeks, unless someone is concerned with changing the default?

francoisfreitag:
I started looking at changing the default. I don’t think it’s worth the hassle. I assume most projects don’t use extensions, and the `INTERNATIONAL` representation is slightly different:

```
>>> str(phonenumbers.format_number(x, phonenumbers.PhoneNumberFormat.INTERNATIONAL))
'+44 20 8366 1177'
>>> str(phonenumbers.format_number(x, phonenumbers.PhoneNumberFormat.E164))
'+442083661177'
```

The added spaces are sure to break code for someone, who doesn’t care about extensions at all. Unless there’s strong support for changing the default, it’ll stay as `E164`.

PR #630, opened 2025-02-14
Support Django 5.2

adamchainz:
Now in alpha: https://www.djangoproject.com/weblog/2025/jan/16/django-52-alpha-1-released/

Some of the test assertions changed in 9b95e59158e8c9baf64f96906787230b64407eb9 needed their version updating because they were due to changes in Django 5.2, rather than 6.0.

Other than that, no new warnings or anything to deal with.


francoisfreitag (approved):
Thank you! :star:

PR #632, opened 2025-03-21
Allow configuring empty_value for SplitPhoneNumberField

ilkkao:
I'm not sure at all if this my mistake or a real issue. But thought it's easiest to explain it as a PR.

Without this fix, in my django app, a string `[]` gets saved to the `phone_number` model view when nothing is entered to the form SplitPhoneNumberField. Model field is defined as:

```
    phone_number = PhoneNumberField(blank=True, null=True)
```

Field in my form:

```
    phone_number = AdminSplitPhoneNumberField(
        required=False,
        help_text='Phone number, country code separated'
    )
```

In my view I just call

```
self.form_valid(my_form)
```

### DIFF
--- a/phonenumber_field/formfields.py
+++ b/phonenumber_field/formfields.py
@@ -105,7 +105,14 @@ class SplitPhoneNumberField(MultiValueField):
     widget = widgets.PhoneNumberPrefixWidget
 
     def __init__(
-        self, *, initial=None, region=None, widget=None, empty_value="", **kwargs
+        self,
+        *,
+        initial=None,
+        region=None,
+        widget=None,
+        empty_value="",
+        max_length=None,
+        **kwargs,
     ):
         """
         :keyword list initial: A two-elements iterable:
@@ -125,6 +132,8 @@ class SplitPhoneNumberField(MultiValueField):
         :keyword ~django.forms.MultiWidget widget: defaults to
             :class:`~phonenumber_field.widgets.PhoneNumberPrefixWidget`
         :keyword empty_value: value to use when the field is empty
+        :keyword int max_length: maximum length of the phone number, when
+            represented as :setting:`PHONENUMBER_DB_FORMAT`.
         """
         validate_region(region)
         region = region or getattr(settings, "PHONENUMBER_DEFAULT_REGION", None)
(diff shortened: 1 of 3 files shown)


phonenumber_field/formfields.py
```diff
@@ -162,7 +162,7 @@ def invalid_error_message(self):
 
     def compress(self, data_list):
         if not data_list:
-            return data_list
+            return None if isinstance(data_list, list) else data_list
```
ilkkao, line 165: or just `return None`

francoisfreitag:
I've seen your message only recently, I'll investigate this issue.

francoisfreitag:
Investigated the issue. From the commit message:

> Previous behavior was to return an empty list for no data, which is surprising and cannot be saved to a django.db.models.CharField in the database.
> Following django.forms.CharField behavior, accepts an empty_value, defaulting to the empty string.

@ilkkao, when you have the chance, could you try out this branch in your project and confirm it solves your issue?

Since your field is nullable, you might need to subclass the `SplitPhoneNumberField` to specify `empty_value=None`. Django has some logic to set the empty value to `None` when the field is nullable, so it might just work out of the box, depending on your project.
https://github.com/django/django/blob/6df19412aabb7d969f5eab4b2ff41269de89b233/django/db/models/fields/__init__.py#L1308

ilkkao:
I'll add a reminder for myself to comment in couple weeks. I'm travelling currently.

francoisfreitag:
Thanks!
I noticed the `SplitPhoneNumberField` is incompatible for use with `formfield` for the model fields, because the model field specifies a `max_length` and the form field does not know of it. I’m working on adding `max_length` to the `SplitPhoneNumberField`.

francoisfreitag:
Thanks for testing! No worries, we all have lives outside of open-source :)
Just to make sure I understand your answer correctly, is this the behavior you expected?
I modeled the `empty_value` normalization after Django:
https://github.com/django/django/blob/adc80dd81eb70d8e11230b9603b025f0d98e1373/django/forms/fields.py#L278

francoisfreitag:
Great to hear, I’ll merge and release this soon.

PR #643, opened 2025-09-23
Add Hungarian locale

karolyi:
- Remove fuzzy markers from de locale messages for national/international phone number validation in form fields.
- Refine German translations for phone number input prompts to include international prefix examples.
- Introduce new hu locale django.po with translations for form fields, model fields, serializer fields, and validators for phone numbers.

### DIFF
--- a/phonenumber_field/locale/hu/LC_MESSAGES/django.po
+++ b/phonenumber_field/locale/hu/LC_MESSAGES/django.po
@@ -1,0 +1,47 @@
+# This file is distributed under the same license as the PACKAGE package.
+#
+msgid ""
+msgstr ""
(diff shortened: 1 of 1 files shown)


francoisfreitag:
Thanks! An [earlier PR](https://github.com/stefanfoulis/django-phonenumber-field/pull/641) was addressing more locales. I merged it, then rebased your branch on it.
The addition of the Hungarian locale remains, can you double check both German and Hungarian locales look correct to you?

francoisfreitag:
> Just a humble advice, you want to use --add-location file when extracting messages, so the .po files wouldn't change with every code line skewing, as the line numbers wouldn't be there.

Thanks for the advice, proposed in #644.

